PR #6, opened 2021-11-10
Rebar: Run Dialyzer on testsuites

dumbbell:
In order to run Dialyzer on testsuites, we have to add `test` to the source directories, even though this is semantically incorrect. We also need to declare PropEr as a regular dependency, not a test-specific one.

The downsides are that PropEr becomes a regular dependency (even when this library is used as a dependency and never tested) and test modules are compiled by default...

Once and if Rebar grows support for Dialyzer on tests, we can get rid of all this.


rebar.config
```diff
@@ -2,19 +2,37 @@
...
+%% In order to run Dialyzer on testsuites, we have to add `test` to the source
+%% directories, even though this is semantically incorrect.
+%%
+%% The downside is that PropEr becomes a regular dependency (even when this
+%% library is used as a dependency and never tested) and test modules are
+%% compiled by default...
+%%
+%% Once and if Rebar grows support for Dialyzer on tests, we can get rid of all
+%% this.
+{extra_src_dirs, ["test"]}.
```
gomoripeti, line 34: would it be an alternative to run dialyzer in the test profile (`rebar3 as test dialyzer`) and only set `{extra_src_dirs, ["test"]}.` in the test profile? This is to avoid adding proper as a regular dep. (sorry to comment from out of the blue, I might be missing the context. Just seen this change in the khepri release notes and it surprised me)

dumbbell: Don't be sorry, your comment is welcome!

That's a very good suggestion. I'm new to Rebar as I mostly use Erlang.mk in the context of RabbitMQ. I like your idea so that PropEr can be treated as a test dependency only (I hated to have to add it to regular runtime deps...).

Do you know if there is a way to set a default profile for Dialyzer? Or do I have to type `as test` everytime I want to run Dialyzer on testsuites?

gomoripeti: I think there is no way to set a default profile for an existing command like dialyzer.

What is even worse is that there is no way to set a profile for an alias either. (So I cannot define something like `{alias, [{dialyzer_with_tests, [{as, test, dialyzer}}]}]}`. 

I just noticed that there is a `check` alias defined. I don't see any way to run the `as test dialyzer` as part of check, only to run the whole check command in the test profile. This is a blocker, looks like the current solution is the only one until some improvements in rebar3.

dumbbell: Hi @gomoripeti! I wanted to let you know that I removed PropEr as a regular dependency (#23). CI sets the `test` profile for Dialyzer. When I run it locally, either I remember about it or CI will shout :-)

PR #30, opened 2022-01-10
khepri_fun: Support binary handling instructions

dcorbacho:


src/khepri_tx.erl
```diff
@@ -268,6 +268,18 @@ ensure_instruction_is_permitted({trim, _, _}) ->
...
+ensure_instruction_is_permitted({bs_init2, _, _, _, _, _, _}) ->
+    ok;
+ensure_instruction_is_permitted({bs_put_string, _, _}) ->
+    ok;
+ensure_instruction_is_permitted({bs_put_binary, _, _, _, _, _}) ->
+    ok;
+ensure_instruction_is_permitted({bs_put_integer, _, _, _, _, _}) ->
+    ok;
+ensure_instruction_is_permitted({bs_append, _, _, _, _, _, _, _, _}) ->
+    ok;
```
dumbbell, line 282: Could you please sort the clauses alphabetically as well?

src/khepri_fun.erl
```diff
@@ -916,22 +925,41 @@ pass2_process_instruction(
...
+              (FieldFlag band 16#04) == 16#04;
+          (native) ->
+              (FieldFlag band 16#10) == 16#10
+      end, [signed, little, native]).
+
+replace_label(Instruction, Pos, #state{mfa_in_progress = {Module, _, _},
+                                       label_map = LabelMap}) ->
+    case element(Pos, Instruction) of
+        {f, 0} ->
+            %% Exception label, for branching
```
dumbbell, line 956: I think we should expand the explanation in the comment and add link to https://github.com/erlang/otp/blob/master/lib/compiler/src/beam_validator.erl#L26-L32 (but a link to a specific commit or tag, so that it remains valid long after the code changes in the `master` branch).

dumbbell (changes requested):

PR #38, opened 2022-02-02
rewrite case labels in select_tuple_arity instruction

the-mikedavis:
:wave: hello!

The disasm scanning to lint transaction functions is super cool!

I noticed the select_tuple_arity instruction doesn't get its cases' labels rewritten, so code that has a case statement branching on the arities of tuples will make the `compile:forms/2` call error out with a message talking about undefined labels. That instruction has the same shape as select_val, so I have it re-using that function clause to rewrite the labels.

What do you think?

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -1111,9 +1111,10 @@ pass2_process_instruction(
   {loop_rec, _, _} = Instruction, State) ->
     replace_label(Instruction, 2, State);
 pass2_process_instruction(
-  {select_val, _, _, {list, Cases}} = Instruction,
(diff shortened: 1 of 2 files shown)


dumbbell:
Thank you very much for this patch! I agree, those two instructions only differ in their first argument and we don't have to interpret it in any way.

Dialyzer will complain about the case clauses which will never match in your testcase. Could you please add your testcase in the `-dialyzer` attribute at the beginning of `tx_funs.erl` under `{no_match, [...]}` (I think it will be `no_match`)? You can test that locally by calling `rebar3 as test dialyzer`.

dumbbell (approved):

dumbbell:
Thank you!

PR #40, opened 2022-02-02
patch more incorrectly decoded bitstring-matching instructions

the-mikedavis:
builds on #39 to decode more bitstring-matching instructions that are not decoded properly by beam_disasm

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -597,7 +597,9 @@ pass1_process_instructions(
 pass1_process_instructions(
   [{BsGetSomething, Ctx, Dst, {u, Live}} | Rest],
   State,
-  Result) when BsGetSomething =:= bs_get_position orelse BsGetSomething =:= bs_get_tail ->
+  Result)
+  when BsGetSomething =:= bs_get_position orelse
+       BsGetSomething =:= bs_get_tail ->
     %% `beam_disasm' did not decode this instruction correctly. We need to
     %% patch it to store `Live' as an integer.
     Instruction = {BsGetSomething, Ctx, Dst, Live},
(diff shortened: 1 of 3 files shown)


src/khepri_fun.erl
```diff
@@ -571,18 +571,37 @@ pass1_process_instructions(
...
     %% flags.
     FieldFlags = decode_field_flags(FieldFlags0),
-    Instruction = {test, bs_get_integer2,
+    Instruction = {test, BsGetSomething,
                    Fail, Live, [Ctx, Size, Unit, FieldFlags], Dst},
     pass1_process_instructions([Instruction | Rest], State, Result);
+pass1_process_instructions(
+  [{BsGetSomething, Ctx, Dst, {u, Live}} | Rest],
+  State,
+  Result) when BsGetSomething =:= bs_get_position orelse BsGetSomething =:= bs_get_tail ->
```
dumbbell, line 600: Could you please wrap this line to stay under the 80-column limit?

test/tx_funs.erl
```diff
@@ -213,6 +213,15 @@ allowed_bs_match_test() ->
            matches_type(queue, proplists:get_value('apply-to', List))
        end).
 
+parse_date(<<Year:4/bytes, $-, Month:2/bytes, $-, Day:2/bytes, _Rest/binary>>) ->
```
dumbbell, line 216: Likewise, please split this line.

dumbbell (changes requested):
Awesome work!

I just made two style-related comment, but otherise the patch looks good to me.

dumbbell:
We plan to upstream those fixes to Erlang at some point (when we have enough of them). That said, this code will stay in Khepri as well to remain compatible with several versions of `beam_disasm`.

dumbbell (approved):

dumbbell:
Thank you!

PR #43, opened 2022-02-03
Add documentation provider

vkatsuba:
For case when last version(`7.0.1`) of project plugin `rebar3_hex` is used for publication of documentation to `hex.pm`, can be received second warning:
```sh
$ rebar3 hex publish docs
...
===> Verifying dependencies...
===> docs argument given, will not publish package
===> No documentation provider was configured, docs will not be generated
```
For this case was added provider as suggested in `rebar3_hex` documentation.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -45,3 +45,5 @@
 {profiles,
  [{test,
    [{deps, [{proper, "1.4.0"}]}]}]}.
+
+{hex, [{doc, edoc}]}.


rebar.config
```diff
@@ -6,7 +6,7 @@
 
 {project_plugins, [rebar3_proper,
                    rebar3_hex,
-                   {rebar3_edoc_extensions, "1.0.0"}]}.
+                   rebar3_edoc_extensions]}.
```
dumbbell, line 9: Could you please remove this unrelated change?

dumbbell (changes requested):
Could you please add the explanation you wrote in the pull request to the commit message too? This way the commit history is self-explanatory; no need to go to GitHub.

vkatsuba:
> Could you please add the explanation you wrote in the pull request to the commit message too? This way the commit history is self-explanatory; no need to go to GitHub.

Done.

rebar.config
vkatsuba: Done. Do I need move those changes in separate commit or open new PR for it? Or for now this is not necessary at all?

dumbbell (approved):

dumbbell:
Thank you!

PR #46, opened 2022-02-07
add translation for 'raise' BIF

the-mikedavis:
See [this](https://github.com/erlang/otp/blob/df48c260e74c3e9058ff8681ce9f554e6fa0fe34/lib/compiler/src/beam_asm.erl#L360-L361) line in beam_asm.

The `raise` instruction needs to be translated to a `bif`. Luckily
the arguments line up very well so it's an easy translation. It was
a bit tough writing a test case that the compiler wouldn't optimize
the raise/try out of, I ended up using the send+receive technique from
a few test cases down.

### DIFF
--- a/src/khepri_tx.erl
+++ b/src/khepri_tx.erl
@@ -300,6 +300,12 @@ ensure_instruction_is_permitted({test_heap, _, _}) ->
     ok;
 ensure_instruction_is_permitted({trim, _, _}) ->
     ok;
+ensure_instruction_is_permitted({'try', _, _}) ->
+    ok;
+ensure_instruction_is_permitted({try_end, _}) ->
+    ok;
+ensure_instruction_is_permitted({try_case, _}) ->
+    ok;
 ensure_instruction_is_permitted(Unknown) ->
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you for your ongoing contributions @the-mikedavis 👍 👍

src/khepri_tx.erl
```diff
@@ -298,6 +300,12 @@ ensure_instruction_is_permitted({test_heap, _, _}) ->
     ok;
 ensure_instruction_is_permitted({trim, _, _}) ->
     ok;
+ensure_instruction_is_permitted({'try', _, _}) ->
+    ok;
+ensure_instruction_is_permitted({try_end, _}) ->
+    ok;
+ensure_instruction_is_permitted({try_case, _}) ->
+    ok;
```
dumbbell, line 308: Could you please split your commit into two and make a separate commit for this chunk?

dumbbell (changes requested):
This looks good to me. I made a small comment to clarify the commit history.

PR #45, opened 2022-02-04
guess var_info for no_bs_start_match2 compilation errors

the-mikedavis:
Given some code like the following:

    trim_leading_dash1(<<$-, Rest/binary>>) -> trim_leading_dash1(Rest);
    trim_leading_dash1(Binary)              -> Binary.

The compiler will add an annotation for the {x,0} register

    {'%',{var_info,{x,0},[accepts_match_context]}}.

This annotation is not available in the disasm, so when we recompile
this function, the compiler gives an error.

This change detects which x-register should accept the match context
by peeking the first instruction after the entrypoint label. This
instruction can either be a {test,bs_start_match3,..} instruction or
bs_start_match4 (if the compiler can determine that the function will
_always_ be passed a binary), both of which contain the register
which will be matched.

closes #44
see also #36

PR #49, opened 2022-02-09
preemptively infer var_info annotations

### LINKED ISSUE #44: guess the `accepts_match_context` `var_info` annotation upon `no_bs_start_match2` failure
Testing more bitstring matching on top of #40, we can write a test case like

```erl
trim_leading_dash(<<$-, Rest/binary>>) ->  trim_leading_dash(Rest);
trim_leading_dash(Binary) ->               Binary.

allowed_bs_match_accepts_match_context_test() ->
    ?assertStandaloneFun(
       begin
           <<"5">> = trim_leading_dash(<<"-5">>)
       end).
```

<details><summary>That fails with this error message...</summary>

```
Failures:

  1) tx_funs:allowed_bs_match_accepts_match_context_test/0
     Failure/Error: {throw,
                     {invalid_tx_fun,
                      {compilation_failure,
                       {error,
                        [{"kfun__tx_funs__-allowed_bs_match_accepts_match_context_test/0-fun-0-__86701401",
                          [{1,beam_validator,
                            {{'kfun__tx_funs__-allowed_bs_match_accepts_match_context_test/0-fun-0-__86701401',
                              tx_funs__trim_leading_dash,1},
                             {{call_only,1,{f,5}},6,no_bs_start_match2}}}]}],
                        []},
                       {'kfun__tx_funs__-allowed_bs_match_accepts_match_context_test/0-fun-0-__86701401',
                        [{run,0}],
                        [],
                        [{function,run,0,2,
                          [{label,1},
                           {func_info,
                            {atom,
                             'kfun__tx_funs__-allowed_bs_match_accepts_match_context_test/0-fun-0-__86701401'},
                            {atom,run},
                            0},
                           {label,2},
                           {allocate,0,0},
                           {move,{literal,<<"-5">>},{x,0}},
                           {call,1,{f,5}},
                           {test,is_eq_exact,{f,3},[{x,0},{literal,<<"5">>}]},
                           {deallocate,0},
                           return,
                           {label,3},
                           {badmatch,{x,0}}]},
                         {function,tx_funs__trim_leading_dash,1,5,
                          [{label,4},
                           {func_info,
                            {atom,
                             'kfun__tx_funs__-allowed_bs_match_accepts_match_context_test/0-fun-0-__86701401'},
                            {atom,tx_funs__trim_leading_dash},
                            1},
                           {label,5},
                           {bs_start_match4,{atom,no_fail},1,{x,0},{x,0}},
                           {test,bs_match_string,
                            {f,6},
                            [{x,0},8,{string,<<"-">>}]},
                           {call_only,1,{f,5}},
                           {label,6},
                           {bs_get_tail,{x,0},{x,0},1},
                           return]}],
                        7}}},
                     [{khepri_tx,to_standalone_fun,2,
                       [{file,
                         "/home/michael/dev-tools/code/khepri/src/khepri_tx.erl"},
                        {line,190}]},
                      {tx_funs,allowed_bs_match_accepts_match_context_test,0,
                       [{file,
                         "/home/michael/dev-tools/code/khepri/test/tx_funs.erl"},
                        {line,233}]},
                      {eunit_test,'-mf_wrapper/2-fun-0-',2,
                       [{file,"eunit_test.erl"},{line,273}]},
                      {eunit_test,run_testfun,1,
                       [{file,"eunit_test.erl"},{line,71}]},
                      {eunit_proc,run_test,1,
                       [{file,"eunit_proc.erl"},{line,531}]},
                      {eunit_proc,with_timeout,3,
                       [{file,"eunit_proc.erl"},{line,356}]},
                      {eunit_proc,handle_test,2,
                       [{file,"eunit_proc.erl"},{line,514}]},
                      {eunit_proc,tests_inorder,3,
                       [{file,"eunit_proc.erl"},{line,456}]}]}
```

</details>

Looking through the compiler I see [this case statement](https://github.com/erlang/otp/blob/9032dc5817eb55f1556061955dab2e9b6143e5c1/lib/compiler/src/beam_validator.erl#L1764-L1767) that I think is throwing the error, and [this test case](https://github.com/erlang/otp/blob/6a48cc051452c756c15659427cd7637e5a94c91d/lib/compiler/test/beam_validator_SUITE.erl#L813) that shows an asm comment annotating the variable with `accepts_match_context`:

```erl
{'%', {var_info, {x, 0}, [accepts_match_context]}}
```

If we put that comment in the function's code right after its entrypoint label, I get a good compile.

I think it makes sense to handle this like in #36, but I'm not sure there's enough information to guess which of the x registers should get the annotation :thinking:

What do you think is a good way to handle this?

the-mikedavis:
supersedes #45
closes #44 

This PR:

- moves the annotation-guessing for tuples and maps into `pass1_process_instructions/1`
- guesses `accepts_match_context` annotations for `bs_start_match*` instructions
- refactors the add_comment_* functionality to add annotations at exact indices within instruction lists upon `compile:forms/2` failure

This is still WIP while I figure out a good way to handle the hard-coded register :thinking:

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -74,7 +74,7 @@
                              to_standalone_arg/2,
                              handle_compilation_error/2,
                              handle_validation_error/3,
-                             add_comment_and_retry/4,
+                             add_comment_and_retry/5,
                              add_comment_to_function/5,
                              add_comment_to_code/3,
                              add_comment_to_code/4,
@@ -305,40 +305,45 @@ handle_validation_error(
    {{Call, 1 = _Arity, {f, EntryLabel}},
     _,
     no_bs_start_match2}},
-  _Error) when Call =:= call orelse Call =:= call_only ->
+  Error) when Call =:= call orelse Call =:= call_only ->
     %% TODO: hard-coded register
     Comment = {'%', {var_info, {x, 0}, [accepts_match_context]}},
     Location = {'after', {label, EntryLabel}},
-    add_comment_and_retry(Asm, EntryLabel, Location, Comment);
+    add_comment_and_retry(Asm, Error, EntryLabel, Location, Comment);
 handle_validation_error(
   Asm,
   {FailingFun,
    {{bs_start_match4, _Fail, _, Var, Var} = FailingInstruction,
     _,
     {bad_type, {needed, NeededType}, {actual, any}}}},
-  _Error) ->
+  Error) ->
     VarInfo = {var_info, Var, [{type, NeededType}]},
     Comment = {'%', VarInfo},
     Location = {before, FailingInstruction},
-    add_comment_and_retry(Asm, FailingFun, Location, Comment);
+    add_comment_and_retry(Asm, Error, FailingFun, Location, Comment);
 handle_validation_error(Asm, _ValidationFailure, Error) ->
     throw({compilation_failure, Error, Asm}).
 
 add_comment_and_retry(
-  Asm, FailingFun, Location, Comment) ->
+  Asm, Error, FailingFun, Location, Comment) ->
     {GeneratedModuleName,
      Exports,
      Attributes,
      Functions,
      Labels} = Asm,
-    Functions1 = add_comment_to_function(
-                 Functions, FailingFun, Location, Comment, []),
-    Asm1 = {GeneratedModuleName,
-            Exports,
-            Attributes,
-            Functions1,
-            Labels},
-    compile(Asm1).
+    try
+        Functions1 = add_comment_to_function(
+                     Functions, FailingFun, Location, Comment, []),
+        Asm1 = {GeneratedModuleName,
+                Exports,
+                Attributes,
+                Functions1,
+                Labels},
+        compile(Asm1)
+    catch
+        throw:duplicate_annotations ->
+            throw({compilation_failure, Error, Asm})
+    end.
 
 add_comment_to_function(
   [#function{name = Name, arity = Arity, code = Code} = Function | Rest],
@@ -385,6 +390,7 @@ add_comment_to_code(
   [Instruction | Rest], Location, Comment, Result) ->
     add_comment_to_code(Rest, Location, Comment, [Instruction | Result]).
 
+merge_comments(Comment, Comment) -> throw(duplicate_annotations);
 merge_comments(
   {'%', {var_info, Var, Attributes1}},
   {'%', {var_info, Var, Attributes2}}) ->
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -73,6 +73,7 @@
                              to_standalone_env/1,
                              to_standalone_arg/2,
                              handle_compilation_error/2,
+                             handle_validation_error/3,
                              add_comment_and_retry/4,
                              add_comment_to_function/7,
                              add_comment_to_code/5,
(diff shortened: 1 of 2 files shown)


src/khepri_fun.erl
```diff
@@ -330,55 +346,39 @@ add_comment_and_retry(
...
+    lists:reverse(Result) ++ [Comment1 | Rest];
+add_comment_to_code([Instruction | Rest], Location, Comment, Counter, Result) ->
+    add_comment_to_code(
+        Rest, Location, Comment, Counter + 1, [Instruction | Result]).
+
+merge_comments(
+  {'%', {var_info, Var, Attributes1}},
+  {'%', {var_info, Var, Attributes2}}) ->
+    {'%', {var_info, Var, Attributes1 ++ Attributes2}};
+merge_comments(Comment, _Instruction) -> Comment.
```
the-mikedavis, line 381: I was wrong about the merging behavior of mulitple comments, we do need to do the merging ourselves.

src/khepri_fun.erl
```diff
@@ -283,45 +285,59 @@ handle_compilation_error(
...
 
+handle_validation_error(
+  Asm,
+  {_FailingFun,
+   {{Call, 1 = _Arity, {f, EntryLabel}},
+    _Index,
+    no_bs_start_match2}},
+  _Error) when Call =:= call orelse Call =:= call_only ->
+    %% TODO: hard-coded register
+    Comment = {'%', {var_info, {x, 0}, [accepts_match_context]}},
```
the-mikedavis, line 310: This is currently hard-coded to work for a 1-arity function so that we know which register to annotate. I think to know which register to use we may need to track the typing of functions which might be difficult. I'll think about it some more :thinking:

the-mikedavis:
It seems like a lot of work to try to reconstruct all of the annotations beam_validator needs and I think these annotations end up only being interpreted by beam_validator. Maybe we should propose adding an option to `compile:forms/2` to disable the beam_validator pass?

dumbbell (changes requested):
Thank you for working on this! I submitted more inline comments.

src/khepri_fun.erl
the-mikedavis: I think this is a good idea but I don't think clauses for merging types would end up being used. For example with a pair that has its first and second element accessed, currently we add the annotations:

```erl
{'%',{var_info,{x,0}, [{type,t_tuple,1,false,#{}}]}}.
{get_tuple_element,{x,0},0,{y,2}}.
{'%',{var_info,{x,0}, [{type,t_tuple,2,false,#{}}]}}.
{get_tuple_element,{x,0},1,{x,0}}.
```

In the clauses added to pass1_process_instructions/3.

We could try to merge these annotations together but it seems complicated: the annotation will only be correct within the scope (all instructions between `{label, _}`s) and I'm not sure how it works when variables are moved around (for example `{move, {x, 1}, {x, 0}}`).

If we do run into cases where the validator disagrees with the types, I think it might make sense to have `merge_comments/2` replace the `{type_}` attribute in Attributes2 with the one in Attributes1 (prefer the one beam_validator says is `needed`)

src/khepri_fun.erl
```diff
@@ -283,45 +285,59 @@ handle_compilation_error(
...
+    %% The first three instructions are always
+    %% - an initial label
+    %% - the func_info instruction
+    %% - the function's entry-point label
```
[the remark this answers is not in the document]
the-mikedavis: Hmm good point, I think we can add annotations here by inserting before the `FailingInstruction` in `{FailingFun, {FailingInstruction, Location, Failure}}` instead of using `Location`.

src/khepri_fun.erl
```diff
@@ -330,55 +341,47 @@ add_comment_and_retry(
...
+    MaybeExistingComment = case Placement of
+                               before -> hd(Result);
+                               'after' -> hd(Rest)
+                           end,
+    Comment1 = merge_comments(Comment, MaybeExistingComment),
+    Rest1 = case Placement of
+              before -> [Comment1, Instruction | Rest];
+              'after' -> [Instruction, Comment1 | Rest]
+            end,
+    lists:reverse(Result) ++ Rest1;
```
the-mikedavis, line 375: We don't discard the existing comment if there is one: beam_validator appears to only acknowledge the last annotation for a variable. What do you think, should we discard the existing comment. I don't imagine it would realistically happen but since the behavior is undocumented it could come back to bite us if we don't discard.

dumbbell: Yeah, I think we should merge comments and discard the previous one. As you say, duplicates could be rejected in the future. Also, it makes the analysis of the generated assembly more straightforward. I mean, duplicate comments could be confusing.

src/khepri_fun.erl
dumbbell: Ok, let's wait and revisit this if we hit a problematic situation.

src/khepri_fun.erl
dumbbell, line 375: Another situation which was previously handled by `are_comments_conflicting()` is the case where the comment we add is insufficient to fix the compilation failure: this prevented an finite loop where we would add more of the same comment after each compilation attempt.

Could you please verify that the comment you add is different from an existing comment? If this occurs, the code could throw `the same exception as the one in the last clause of `handle_compilation_error()`.

dumbbell (changes requested):

the-mikedavis:
I think it makes sense to tackle the 'TODO' for the hard-coded register ([this](https://github.com/rabbitmq/khepri/pull/49#discussion_r803024263)) in a separate PR: it will probably be a decently large change on its own. What do you think?

dumbbell (approved):

PR #54, opened 2022-02-16
khepri_fun: Add `select_tuple_arity` type annotation

dcorbacho:
The `select_tuple_arity` is used in the following code for instance:

```erlang
case Tuple of
    {ok, _, _}   -> true;
    {error, _}   -> false;
    {timeout, _} -> false
end
```

In other words, the VM can simply verify the arity of the tuple to decide between `true` and `false` if the atom in the first position was implicitly verified earlier in the code.

`beam_validator` expects the annotation to specify a tuple without any specific arity.


dumbbell (changes requested):
Could you please add a testcase for this annotation?

dumbbell:
I also started to figure out a testcase by the way :-)

dumbbell (approved):

PR #63, opened 2022-03-15
khepri_fun: propagate argument annotations to callee in bs_no_start_match2 validation failures

the-mikedavis:
This is a follow up for removing the hard-coded register `FIXME` (https://github.com/rabbitmq/khepri/pull/49#issuecomment-1039196029). Sorry the title is so long and jargon-y 😅

A simplification of the problem is that function `A` calls function `B` calls function `C`. `A` passes an argument to `B` that `accepts_match_context`, but `B` doesn't have any annotations for its parameters, so the `beam_validator` emits an error on the call instruction saying: `no_bs_start_match2`.

<details><summary>I got a bit derailed looking into this by function <code>C</code>...</summary><br>

In the test case fixtures, `C` has a parameter that `accepts_match_context`, so I thought that in order to fix this, we would need to determine the types of each function in the call graph (which naively seems easy based on [`beam_validator:build_function_table/2`](https://github.com/erlang/otp/blob/59aa9c55b3c7eb7e646c98b8888af43b135e7664/lib/compiler/src/beam_validator.erl#L238-L264)). It turns out though that this is not the case (see the commit message for edf1e98 for some more details on this).

<hr></details>

In order to handle this particular `beam_validator` failure, we detect the annotations within the branch of the failed call instruction and add those annotions to the callee right after its entrypoint.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -80,10 +80,14 @@
                              to_standalone_arg/2,
                              handle_compilation_error/2,
                              handle_validation_error/3,
-                             add_comment_and_retry/5,
-                             add_comment_to_function/5,
-                             add_comment_to_code/3,
-                             add_comment_to_code/4,
+                             add_comments_and_retry/5,
+                             add_comments_to_function/5,
+                             add_comments_to_code/3,
+                             add_comments_to_code/4,
+                             find_comments_in_branch/2,
+                             find_comments_in_branch/4,
+                             split_comments/1,
+                             split_comments/2,
                              merge_comments/2]}).
 
 -type fun_info() :: #{arity => arity(),
@@ -426,17 +430,20 @@ handle_compilation_error(Asm, Error) ->
 
 handle_validation_error(
   Asm,
-  {_FailingFun,
-   {{Call, 1 = _Arity, {f, EntryLabel}},
-    _,
+  {{_, Name, Arity},
+   {{Call, _Arity, {f, EntryLabel}},
+    CallIndex,
     no_bs_start_match2}},
   Error) when Call =:= call orelse Call =:= call_only ->
-    %% FIXME: We currently hard-code the register to annotate because we don't
-    %% have an immediate way to tell which one should be annotated.
-    Register = {x, 0},
-    Comment = {'%', {var_info, Register, [accepts_match_context]}},
+    {_GeneratedModuleName,
+     _Exports,
+     _Attributes,
+     Functions,
+     _Labels} = Asm,
+    #function{code = Instructions} = lookup_function1(Functions, Name, Arity),
+    Comments = find_comments_in_branch(Instructions, CallIndex),
     Location = {'after', {label, EntryLabel}},
-    add_comment_and_retry(Asm, Error, EntryLabel, Location, Comment);
+    add_comments_and_retry(Asm, Error, EntryLabel, Location, Comments);
 handle_validation_error(
   Asm,
   {FailingFun,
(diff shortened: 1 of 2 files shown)


src/khepri_fun.erl
```diff
@@ -445,22 +453,57 @@ handle_validation_error(
...
-    add_comment_and_retry(Asm, Error, FailingFun, Location, Comment);
+    add_comments_and_retry(Asm, Error, FailingFun, Location, Comments);
 handle_validation_error(Asm, _ValidationFailure, Error) ->
     throw({compilation_failure, Error, Asm}).
 
-add_comment_and_retry(
-  Asm, Error, FailingFun, Location, Comment) ->
+find_comments_for_call(
+  [#function{name = Name, arity = Arity, code = Instructions} | _Rest],
+  Name, Arity, CallIndex) ->
```
dumbbell, line 464: Could you please add a comment which explains the logic of `find_comments_for_call()` and `find_comments_in_branch()`? Perhaps with a short example of the assembly.

I find it handy to have comments like this, so that we don't have to go to the commit message to understand what's going on.

src/khepri_fun.erl
```diff
@@ -445,22 +453,57 @@ handle_validation_error(
...
+    [{'%', {var_info, Var, Info}} || {Var, Info} <- maps:to_list(VarInfos)];
+find_comments_in_branch(
+  [{'%', {var_info, Var, Info}} | Rest], CallIndex, Counter, VarInfos) ->
+    VarInfos1 = maps:put(Var, Info, VarInfos),
+    find_comments_in_branch(Rest, CallIndex, Counter + 1, VarInfos1);
+find_comments_in_branch(
+  [{move, Src, Dst} | Rest], CallIndex, Counter, VarInfos) ->
+    VarInfos1 = case VarInfos of
+                    #{Src := Info} -> maps:put(Dst, Info, VarInfos);
+                    _ -> maps:remove(Dst, VarInfos)
```
dumbbell, line 484: I usually align arrows horizontally when the clause body fits on the same line. Do you mind to do that as well?

src/khepri_fun.erl
```diff
@@ -445,22 +453,57 @@ handle_validation_error(
...
+                end,
+    find_comments_in_branch(Rest, CallIndex, Counter + 1, VarInfos1);
+find_comments_in_branch(
+  [{label, _Label} | Rest], CallIndex, Counter, _VarInfos) ->
+    %% Each branch (separated by labels) has idependent typing
```
dumbbell, line 489: Typo in "idependent".

src/khepri_fun.erl
```diff
@@ -445,22 +453,57 @@ handle_validation_error(
...
+    find_comments_in_branch(Rest, CallIndex, Counter + 1, #{});
+find_comments_in_branch(
+  [_Instruction | Rest], CallIndex, Counter, VarInfos) ->
+    find_comments_in_branch(
+      Rest, CallIndex, Counter + 1, VarInfos).
+
+
```
dumbbell, line 496: Could you please keep a single empty line?

dumbbell (changes requested):
Thank you very much! Great patch with detailed commit messages! This is really appreciated.

I made a few comments, mostly nitpicking :-)

src/khepri_fun.erl
the-mikedavis: I ended up refactoring out `find_comments_for_call` since it was really just duplicating `lookup_function1/3` and then calling out to `find_comments_in_branch/2`. It seems more composable to keep that logic in `lookup_function1/3` (but maybe we should rename that function now?)

dumbbell: Good idea!

dumbbell (approved):

PR #67, opened 2022-03-21
khepri_fun: fix coverage for infix operators

the-mikedavis:
The compiler is smart enough to optimize away pretty much all the
instructions that were in `allowed_erlang_expressions_test/0'. In
fact, the asm for the begin/end block in that test came out to:

    {function, '-allowed_erlang_expressions_test/0-fun-0-', 0, 722}.
      {label,721}.
        {line,[{location,"test/tx_funs.erl",65}]}.
        {func_info,{atom,tx_funs},
                   {atom,'-allowed_erlang_expressions_test/0-fun-0-'},
                   0}.
      {label,722}.
        {test,has_map_fields,{f,723},{literal,#{a => b}},{list,[{atom,a}]}}.
        {move,{literal,#{a => c}},{x,0}}.
        return.
      {label,723}.
        {line,[{location,"test/tx_funs.erl",106}]}.
        {badmatch,{literal,#{a => b}}}.

This is a refactor of that test case which treats each case differently
so that the compiler doesn't think any of the calculations can be
discarded.

Also included is a commit for expanding floating point arithmetic
support: a few new instructions and a translation.


test/tx_funs.erl
```diff
@@ -37,6 +36,15 @@
 -define(assertToFunThrow(Expected, Expression),
         ?assertThrow(Expected, ?make_standalone_fun(Expression))).
 
+%% The compiler is smart enough to optimize away many instructions by
+%% inspecting types and values. `mask/1' confuses the compiler by sending
+%% and receiving the value.
+-define(mask(Value),
+        begin
+            self() ! Value,
+            receive __Msg -> __Msg end
```
the-mikedavis, line 45: The compiler gives a warning about duplicated bindings when using the mask macro twice in a scope. I didn't think this would be an issue since `__Msg` is only in-scope for the `receive` block, but it turns out that the binding leaks out of scope:

```erl
1> self() ! 1.
1
2> receive Msg -> Msg end.
1
3> Msg.
1
```

Is there a way of introducing hygienic bindings in a macro in Erlang?

src/khepri_tx.erl
```diff
@@ -256,6 +258,14 @@ ensure_instruction_is_permitted({deallocate, _}) ->
...
 ensure_instruction_is_permitted({func_info, _, _, _}) ->
     ok;
+ensure_instruction_is_permitted({fconv, _, _}) ->
+    ok;
+ensure_instruction_is_permitted(fclearerror) ->
+    ok;
+ensure_instruction_is_permitted({fcheckerror, _}) ->
+    ok;
+ensure_instruction_is_permitted({fmove, _, _}) ->
+    ok;
```
the-mikedavis, line 268: Should I split the commit so the float arithmetic instructions have their own commit and message?

dumbbell:
Thank you! I meant to do that for quite some time but didn't come to it...

src/khepri_tx.erl
dumbbell: Were they required to make the refactored testsuite work?

dumbbell (approved):

PR #76, opened 2022-04-01
Fix how dual licensing is applied

dumbbell:
Now `LICENSE` uses the SPDX syntax to point to the two licenses we use. There are `LICENSE.*` files to remind the content of both licenses.

`khepri.app` is also fixed; it was only mentionning a single license so far.


michaelklishin:
@dumbbell is this a GitHub feature/convention? Should we do this in other (smaller) repos?

dumbbell:
It's not a GitHub convention and GitHub relying on [licensee](https://licensee.github.io/licensee/), this is unsupported (GitHub won't display all the licenses on the right sidebar).

I just searched if there were "official" ways to document licences. I found the following comment speaking about what SPDX suggests:
https://softwareengineering.stackexchange.com/questions/304874/declaring-multiple-licences-in-a-github-project/371435#371435

And SPDX drives many new conventions around licenses, so I thought about following this one too.

Issue #19, opened 2021-12-08
Support `call_fun2` Beam instruction


dumbbell:
While testing Khepri with Erlang master branch (the future Erlang 25), @mkuratczyk saw that Khepri's function extraction and transaction support doesn't support the new `call_fun2` instruction.

Beside while-listing it, we need to study it and determine if Khepri must follow the call and copy/extract more code.

the-mikedavis:
This test case creates a `call_fun2`:

```erl
apply_fun_to_args(Fun, Arg1, Arg2) ->
    Fun(Arg1, Arg2).

allowed_higher_order_external_call_test() ->
    ?assertStandaloneFun(
        begin
            Fun = fun erlang:min/2,
            1 = apply_fun_to_args(Fun, 1, 2)
        end).
```

With a practical example being something like [this from `khepri.erl`](https://github.com/rabbitmq/khepri/blob/62197b4ac85351bef9501d17bb4b0a28c8b041f1/src/khepri.erl#L419).

In asm this looks like:

```erl
{function, apply_fun_to_args, 3, 240}.
  {label,239}.
    {line,[{location,"test/tx_funs.erl",493}]}.
    {func_info,{atom,tx_funs},{atom,apply_fun_to_args},3}.
  {label,240}.
    {'%',{var_info,{x,0},[{type,{t_fun,2,any,any}}]}}.
    {'%',{var_info,{x,1},[{type,{t_integer,{1,1}}}]}}.
    {'%',{var_info,{x,2},[{type,{t_integer,{2,2}}}]}}.
    {allocate,0,1}.
    {move,{integer,2},{x,1}}.
    {move,{x,0},{x,2}}.
    {move,{integer,1},{x,0}}.
    {line,[{location,"test/tx_funs.erl",494}]}.
    {call_fun2,{atom,safe},2,{tr,{x,2},{t_fun,2,any,any}}}.
    {deallocate,0}.
    return.
```

But the disasm looks a bit mangled:

```erl
{call_fun2,{atom,safe},2,{tr,{x,2},{{t_fun,any,any,any},0,18446744073709551615}}
```

I think we should be able to discard the wrapping `{Type, 0, 18446744073709551615}` and just use `Type` but I'm not sure yet where it comes from.

[...]

PR #81, opened 2022-04-06
Support Erlang/OTP 25

dumbbell:
This includes at least the support of the new `bs_create_bin` instruction.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -12,8 +12,7 @@
             warnings_as_errors]}.
 
 {dialyzer, [{plt_extra_apps, [edoc, eunit, inets, mnesia, proper, ssl, xmerl]},
-            {warnings, [race_conditions,
-                        underspecs,
+            {warnings, [underspecs,
                         unknown,
                         unmatched_returns]}]}.
 
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -1159,28 +1159,41 @@ pass1_process_instructions(
   State,
   Result) ->
     State1 = ensure_instruction_is_permitted(Instruction, State),
+    Src1 = fix_type_tagged_beam_register(Src),
+    Instruction1 = setelement(2, Instruction, Src1),
+
+    Reg = get_reg_from_type_tagged_beam_register(Src1),
     Type = {t_tuple, Element + 1, false, #{}},
-    VarInfo = {var_info, Src, [{type, Type}]},
+    VarInfo = {var_info, Reg, [{type, Type}]},
     Comment = {'%', VarInfo},
-    pass1_process_instructions(Rest, State1, [Instruction, Comment | Result]);
+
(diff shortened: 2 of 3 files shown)


dumbbell:
I think the `bs_create_bin` instruction arguments are only partially decoded. In particular, it could have strings among its variable list of arguments. Strings are supposed to have a `{string, Binary}` element, but we only get a `{u, Integer}` from `beam_disasm`. Indeed, the `decode_arg()` function in `beam_disasm` never produces strings.

Perhaps the code is incomplete. I will compare that to `beam_load.c` to understand what's missing. This time, it will probably be difficult to implement a fix in `khepri_fun` because we lack information. We can submit a patch upstream of course, but we might need to consider a fork of that module if the Erlang release calendar doesn't fit our need.

dumbbell:
> Strings are supposed to have a `{string, Binary}` element, but we only get a `{u, Integer}` from `beam_disasm`.

In fact, the instruction's string really has an integer in the beam file in place of a binary. That's because it's an index into a table elsewhere in the beam file. Exactly like the `line` instructions pointing to entries in the "Line" chunk.

dumbbell:
We can't test automatically against Erlang 25 in GitHub Actions for now. There is no prebuilt versions available on repo.hex.pm and the `setup_beam` action doesn't know about Erlang 25 yet.

This explains the drop in code coverage: the new `bs_create_bin` instruction is introduced in Erlang 25 only.

michaelklishin:
@dumbbell should we use a list of tasks to keep track of what else we might want to support?
This PR works great on Erlang 25 (the tip).

the-mikedavis:
The new [`call_fun2/3`](https://github.com/erlang/otp/blob/41f5ee8cf217d10bd3a736a948d5335f591a2b85/lib/compiler/src/genop.tab#L664) instruction might fall under this (#19). I'll see if I can make a test case that can produce it.

dumbbell:
> @dumbbell should we use a list of tasks to keep track of what else we might want to support?

Thank you for testing!

In fact, I was thinking to just merge this current patch, if it works for anyone, not just me. I would feel better if I could run tests against Erlang 25 in CI as well but that won't happen soon, I guess.

After that, Erlang 25 will be officially supported and we will probably need more patches along the road to keep it working. That's the same for previous versions of Erlang: we continue to discover cases where a Beam instruction is unsupported by Khepri, with both Erlang 23 and Erlang 24.

michaelklishin:
@dumbbell SGTM

dumbbell:
> I would feel better if I could run tests against Erlang 25 in CI as well but that won't happen soon, I guess.

In fact, I misread the code of `erlef/setup-beam` and the bot that deploys builds of Erlang to repo.hex.pm. There are snapshots of `main` and `maint` compiled every day and every release tags are published, including release candidates.

I just enabled Erlang 25 using 25.0-rc3 in CI (see #88).

Issue #75, opened 2022-04-01
Implement memory management for the standalone functions cache


dumbbell:
In #72, we introduce a cache for already extracted standalone functions to improve performances.

Currently, this cache will grow indefinitely. We need some kind of policy to clean it up.

the-mikedavis:
What would you think about an ETS-based LRU cache for this? Would you be open to a contribution?

michaelklishin:
In other RabbitMQ projects where we have a cache, it is sometimes an interface that you can implement however you like and configure a module to be used. Maybe it's an overkill for Khepri but there is no single best way to do K/V caching in process, so an interface makes certain sense to me. Our first implementation can be ETS-based :)

dumbbell:
> What would you think about an ETS-based LRU cache for this? Would you be open to a contribution?

Yes, feel free to experiment!

PR #82, opened 2022-04-11
Use an LRU cache for standalone funs

the-mikedavis:
I tried out micro-benchmarking some LRU styles: [the-mikedavis/lru_bench](https://github.com/the-mikedavis/lru_bench/tree/7c65a323d253b66a157ac3a720bfbc8fde431205) and I found one that I think works pretty well, especially for concurrent read/write throughput. I need to plug this in to khepri-benchmark to see how it fares with concurrency but the micro-benchmarks say that this cache is ~1.4x slower for `get/2`s and ~1.8x slower for `put/2`s compared to `persistent_term` which I think is not too bad.

Mostly I had fun beating [lru](https://hex.pm/packages/lru) which is linear runtime (this implementation is logarithmic) on the cache capacity for both `put/2` and `get/2`. Look how it crawls in that XX-Large case! 😄

I'd like to improve the documentation and maybe rename some variables but I thought I'd open this up to discussion earlier rather than later. Plus I have some questions:

* Does this match expectations for interfaces/behaviours? I based it on [`rabbit_auth_cache`](https://github.com/rabbitmq/rabbitmq-server/blob/7abf749a60458aba9d6c9e6bdec1d3b0b2254007/deps/rabbitmq_auth_backend_cache/src/rabbit_auth_cache.erl)
* Should I be tuning any of the read/write concurrency settings in `ets`?

And mostly: is the lack of atomicity in `put/2` and `get/2` ok? A tla+ checker might point out some nasty scenarios from concurrent `put/2`s and `get/2`s, but given that this cache is purely for speed, I think the fidelity of the data doesn't matter much: if keys are dropped or something lives in the cache for longer than expected, it's no big deal. What do you think?

connects #75

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -220,7 +220,7 @@ fun((#{calls := #{Call :: mfa() => true},
                      should_process_function_fun(),
                      is_standalone_fun_still_needed =>
                      is_standalone_fun_still_needed_fun(),
-                     function_cache => module()}.
+                     standalone_fun_cache => module()}.
 %% Options to tune the extraction of an anonymous function.
 %%
 %% <ul>
@@ -456,7 +456,7 @@ get_cached_standalone_fun(
   #state{fun_info = #{module := Module}, options = Options} = State)
   when Module =/= erl_eval ->
     Key = standalone_fun_cache_key(State),
-    FunctionCache = maps:get(function_cache, Options, khepri_fun_cache),
+    FunctionCache = maps:get(standalone_fun_cache, Options, khepri_fun_cache),
     case FunctionCache:get(Key, undefined) of
         #{standalone_fun := StandaloneFunWithoutEnv,
           checksums := Checksums,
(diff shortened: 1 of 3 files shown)


michaelklishin:
Khepri already offers a preference for Ra query types (local, leader). In many cases, a cache miss or getting stale data can be acceptable. In others, a way to bypass the cache would
give the user full control.

the-mikedavis:
Also I updated the micro-benchmark (https://github.com/the-mikedavis/lru_bench/tree/0263ed83bc35b58d2e50a95302c849d666492238) to run in parallel and this implementation seems to scale even better under concurrent access. It's still ~1.5x slower than `persistent_term` for `get/2`s but actually ~2.5x faster for `put/2`s 🚀

michaelklishin:
Ah, I see. Then perfect read consistency is not a goal for this cache since it does not store user data. I trust your judgement @the-mikedavis 👍

the-mikedavis:
Closing for now - we can revisit this later if the memory consumption of standalone funs becomes more pressing. We might end up re-using or modifying the LRU cache from this branch in the query cache.

This needs a bit of work: the cache should take a callback that it executes when evicting an entry so we can unload the standalone func's module. That way we also purge the memory taken up by the code as well.

PR #84, opened 2022-04-20
Fix typo handling deleted nodes on aftermath

dcorbacho:
Added test that reproduces this use case, using an `if_node_exists condition` on `keep_while` that must be exercised


dumbbell (approved):
Thank you!

PR #86, opened 2022-04-20
Handle `{atom,safe}` variant of OTP-25 `call_fun2/3`

the-mikedavis:
This PR only handles the `{atom,safe}` variant: `{atom,unsafe}` appears to only happen in `try` blocks ([example](https://github.com/rabbitmq/khepri/blob/0af07327fcff34c91072448b3df3605f58d41440/src/khepri_fun.erl#L1789)).

Also included is a check for `{move,{literal,Fun},Dst}` instructions that processes any literal functions. I'm a little stuck on how to handle replacing the function with its extracted equivalent (see 3e7c340341ea0cb46ca3eec9ed0d0025ff9ac638). Exporting these extracted functions would be the simplest way but I'm hoping there's a clever way to capture a local function instead so we can keep the current exports (just `run/Arity`).

connects #19


dumbbell:
Thank you for starting to work on this!

Good find about the `{move, {literal, fun tx_funs:min/2}, {x,0}}`! To extract it, you can't use the same state because a state instance is made for a specific anonymous function extraction process. You need to call `to_standalone_fun1(Fun, Options)` (i.e. with the same options as the function being extracted). See the following code as an example:
https://github.com/rabbitmq/khepri/blob/0af07327fcff34c91072448b3df3605f58d41440/src/khepri_fun.erl#L2142-L2153

Now, I need to think about how we should handle this. Perhaps we need to modify the `#standalone_fun{}` record to have a place to store other "attached" extracted functions. `khepri_fun:exec/2` will have to load all attached extracted functions.

It could be tricky if the attached function has a non-empty environment. That said, it's possible that the `{literal, fun ...}` construct will always reference external functions and thus will always have an empty environment (in which case we should assert for that).

The assembly code would be modified to reference the name of the extracted module & function instead:
```erlang
%% Before:
{move, {literal, fun tx_funs:min/2}, {x, 0}}.

%% After:
{move, {literal, fun kfun_long_module_name_with_some_hash:run/2}, {x, 0}}.
```

dumbbell:
Also, this should probably be two different patches in the end :-)

dumbbell:
I opened #101 to address literal funs in `move` instructions.

Issue #92, opened 2022-04-25
Accept binaries as Unix-like paths


dumbbell:
Currently a Khepri path is either a native path (a list of atoms, binaries and conditions) or a Unix-like path which is an Erlang string.

Several other Beam languages such as Elixir or Gleam implement strings as Erlang binaries. Therefore, it would be nice to extend Unix-like path to accept binaries as well. This should be even simpler to parse because we don't have to determine if the passed list is a native path or a Unix-like string!

The easy part is to update `khepri_path:from_string/1` to accept and parse binaries. The boring part is to update all function specs... A new type should be introduced to include `khepri_path:path()`, `string()` and `binary()`.

the-mikedavis:
I see a TODO for changing that `path/0` to `node_path/0`: https://github.com/rabbitmq/khepri/blob/686a16012b98b2baf6f71e56da534ae143d30de9/src/khepri_path.erl#L83-L91

What would you think about types like these?

```erl
%% (renamed from `path/0`)
-type node_path() :: [component()].
%% A native path to a node.

-type unix_path() :: string() | binary().
%% A path as a string or binary.

-type path() :: unix_path() | node_path().
%% A path to a node. Either a string, binary, or a native path.
```

michaelklishin:
What constitutes a "native pattern"? I assume a UNIX pattern is slash-separated? If so, `slash_separated` or `fs_like_path` could be slightly more descriptive?

the-mikedavis:
If I understand it correctly, native paths are the internal representation for paths

```erl
[stock, wood, <<"oak">>].
```

and then unix paths are a convenient string-based syntax that can be translated into native paths

```erl
khepri_path:from_string("/:stock/:wood/oak").
%%=> [stock, wood, <<"oak">>]
```

With patterns being a supertype that also allows conditions from `khepri_condition`:

```erl
%% native pattern:
[stock,
 #if_all{conditions = [wood,
                       #if_node_exists{exists = true}]},
 oak].
```

I like the connotations of "unix" and "native" but unix isn't totally perfect because the `:atom` syntax doesn't have an equivalent in unix paths. Maybe there's a good word that describes unix paths as a convenient syntax (similar to "native" implying that native paths are the internal representation)?

michaelklishin:
So they are list paths and slash-separated or filesystem-like paths.

dumbbell:
I created a separate pull request (#94) to handle the redefinition of the `path()` type and how things should be named.

PR #94, opened 2022-05-02
Redefine the `path()` type

dumbbell:
The main improvement of this branch is the redefinition of the `path()` and `pattern()` types. We are now explicit on which types of path are accepted in which places.

The types are:
* `native_path()`, the internal representation made of atoms, binaries and, for patterns only, conditions.
* `unix_path()`, the Unix-like path representation, which can be either an Erlang string (list of Unicode characters) or an Erlang binary (which can be any UTF encoding).

The `path()` type is now the union of both `native_path()` and `unix_path()`.

Likewise for path patterns with `native_pattern()`, `unix_pattern()` and `pattern()`. Note that in the case of patterns, only a few conditions can be represented in the Unix-like path pattern.

See #92 for the discusion which triggered this change.

### DIFF
--- a/src/khepri_path.erl
+++ b/src/khepri_path.erl
@@ -32,25 +32,9 @@
 %%                oak].
 %% '''
 %%
-%% To be user-friendly, Unix-like string-based paths are accepted by most
-%% functions. These <em>Unix paths</em> have the following syntax:
-%%
-%% <ul>
-%% <li>Path components are separated by a forward slash, `/'.</li>
-%% <li>Atom-based node IDs are prefixed with a `:' character: `:wood'.</li>
-%% <li>Binary-based node IDS are written as-is: `oak'.</li>
-%% <li>Atom and binaries can be percent-encoded.</li>
-%% <li>An absolute path must start with `/', otherwise it is considered a
-%% relative path</li>
-%% <li>`.' and `..' represent `?THIS_NODE' and `?PARENT_NODE'
-%% respectively</li>
-%% <li>Simple glob patterns are accepted:
-%% <ul>
-%% <li>`abc*def' is the same as `#if_name_matches{regex = "^abc.*def$"}'</li>
-%% <li>`*' is the same as `?STAR' or `#if_name_matches{regex = any}'</li>
-%% <li>`**' is the same as `?STAR_STAR' or `if_path_matches{regex = any}'</li>
-%% </ul></li>
-%% </ul>
+%% To be user-friendly, string-based and binary-based <em>Unix-like paths</em>
+%% are accepted by most functions. The syntax of these <em>Unix paths</em> is
+%% described in the {@link unix_path()} type documentation. Example:
 %%
 %% ```
 %% %% Unix path, equivalent of the first native path example.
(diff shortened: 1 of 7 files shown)


dumbbell:
What is not reflected in the patch is the discussion about the naming of the types. Let me copy-paste the relevant comments after I asked feedback on this "native" and "Unix-like" path naming:

Me:
> What do you think? Do you suggest better names than "native" or "Unix"?

@michaelklishin:
> What constitutes a "native pattern"? I assume a UNIX pattern is slash-separated? If so, `slash_separated` or `fs_like_path` could be slightly more descriptive?

@the-mikedavis:
> [gives an explanation of what a native path/pattern is]
>
> I like the connotations of "unix" and "native" but unix isn't totally perfect because the `:atom` syntax doesn't have an equivalent in unix paths. Maybe there's a good word that describes unix paths as a convenient syntax (similar to "native" implying that native paths are the internal representation)?

@michaelklishin:
> So they are list paths and slash-separated or filesystem-like paths.

dumbbell:
@the-mikedavis said:
> I like the connotations of "unix" and "native" but unix isn't totally perfect because the `:atom` syntax doesn't have an equivalent in unix paths. Maybe there's a good word that describes unix paths as a convenient syntax (similar to "native" implying that native paths are the internal representation)?

Unix paths in modern implementations assume that filenames are valid Unicode, so `:atom` is a valid filename. Older implementations assuming 7-bit US-ASCII would accept it too. Even newlines are just fine, not that they are a good idea :-) I should say "conventions" rather than "implementation" because the actual implementations are very permissive. I think only the `/` character is actually reserved. `:atom` is not accepted on Windows however AFAICT, because it is used to separate the drive letter from the path.

@kjnilsson was inspired by ZooKeeper for this hierarchical organization of data in Khepri. Here is what ZooKeeper's overview says:
> The name space provided by ZooKeeper is much like that of a standard file system. A name is a sequence of path elements separated by a slash ("/"). Every znode in ZooKeeper's name space is identified by a path. And every znode has a parent whose path is a prefix of the znode with one less element; the exception to this rule is root ("/") which has no parent. Also, exactly like standard file systems, a znode cannot be deleted if it has any children.

They don't say "Unix-like", they say "standard file system" path. I don't like this paragraph because there is no such thing as a "standard file system". The way we write paths like "/path/to/something" or "\path\to\something" has nothing to do with filesystems in the first place. Filesystems don't care about how file and directory names are separated in a string. They may have reserved characters inside filenames, but that's it. Only layers above file systems are interested in the path syntax.

Unix systems use forward slash only. Windows accepts both forward- and backslashes. That's why I went with "Unix-like" to express the fact that Khepri expects path components to be separated by `/` in string-based paths.

As for "native path", I used this word because it doesn't require any processing to be used by Khepri. As @the-mikedavis said, that's both the public and internal representation of paths. That's unlike the "Unix-like path" which requires parsing and currently can't represent all conditions in patterns.

the-mikedavis:
I like "native" and "unix": I think they're intuitive.

With the `:atom` syntax part I just meant that the syntax isn't special in unix paths - it's just a filename as you say rather than something represented differently internally. Plus I like that syntax coming from Elixir :)

I would prefer "unix" or "unix-like" to "standard file system." Besides the backslashes, it's a bit of a mouthful.

I agree with removing the TODO around `node_path/0`: I think the type names in this PR are good. `khepri_path:path/0` being a type expressing a path to a Khepri node makes sense to me 👍

src/khepri_path.erl
```diff
@@ -87,30 +87,60 @@
...
+%% A native path is a list of atoms, binaries and special components.
+
+-type native_pattern() :: [pattern_component()].
+%% Path pattern which may match zero, one or more nodes.
+%%
+%% A native pattern is a list of atoms, binaries, special components and
+%% conditions.
+
+-type unix_path() :: string() | binary().
+%% Unix-like path to a node.
```
the-mikedavis, line 102: What would you think about moving some of the moduledoc's explanation on unix paths down to this type? On the one hand it's nice to get the full overview by reading the moduledoc but on the other I would expect to find the full explanation of the unix path syntax here with `unix_pattern/0` explaining the added glob syntax

dumbbell: Good idea. I expanded the documentation and moved some things in the types' description as you suggested.

michaelklishin:
"native" and "unix" sound OK to me 👍

dumbbell:
Thank you @michaelklishin and @the-mikedavis for your feedback!

PR #99, opened 2022-05-03
Add Elixir sigils for `khepri_path:path/0`

the-mikedavis:
Similar to #98 and based on discussion in the [ErlangForums thread](https://erlangforums.com/t/khepri-a-tree-like-replicated-on-disk-database-library-for-erlang-and-elixir-introduction-feedbacks/438/44?u=the-mikedavis), this PR explores using Elixir macros in an Erlang dependency.

To be clear: I don't think this PR should be merged. As it turns out, it's pretty simple to do this in pure Erlang since sigils are either regular functions or Elixir macros:

```erl
sigil_p(Path, _Opts) ->
    from_string(Path).

sigil_P(Path, _Opts) ->
	from_string(Path).
```

```elixir
iex> import :khepri_path
iex> ~p"/stock/wood/:oak"
["stock", "wood", :oak]
iex> ~P"/stock/wood/:oak"
["stock", "wood", :oak]
```

(8e9483a has notes on the difference between lowercase and uppercase.) I'll follow up with a PR to add those two functions if you'd like 🙂

But this was a fun exploration into including Elixir files in Erlang dependencies without requiring that Erlang users install Elixir/Mix. The main pitfall I ran into is that documentation isn't combined for the Erlang API and the Elixir API: you would have to generate Erlang documentation separately from Elixir documentation. This could be an interesting issue to bring up in ex_doc.

Overall, I don't think Elixir-specific bindings are necessary to include in Khepri, but this could be a cool approach for an Erlang library that has an API based on parse-transforms: the corresponding Elixir modules could use Elixir macros instead.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,12 +6,18 @@ on:
 
 jobs:
   Test:
-    name: Test on Erlang/OTP ${{ matrix.otp_version }} and ${{ matrix.os }}
+    name: Test on Erlang/OTP ${{ matrix.beam_versions.otp }}, Elixir ${{ matrix.beam_versions.elixir }} and ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
-        otp_version: [23, 24, '25.0-rc3']
+        beam_versions:
+          - otp: 23
+            elixir: '1.11.4'
+          - otp: 24
+            elixir: '1.13.4'
+          # - otp: '25.0-rc3'
+          #   elixir: '1.13.4'
         os: [ubuntu-latest, windows-latest]
 
     env:
@@ -22,12 +28,13 @@ jobs:
       - uses: erlef/setup-beam@v1
         id: install-erlang
         with:
-          otp-version: ${{matrix.otp_version}}
+          otp-version: ${{ matrix.beam_versions.otp }}
+          elixir-version: ${{ matrix.beam_versions.elixir }}
           rebar3-version: '3.18.0'
 
       - name: Restore Dialyzer PLT files from cache
         uses: actions/cache@v2
-        if: ${{ matrix.otp_version == env.RUN_DIALYZER_ON_OTP_RELEASE && matrix.os == 'ubuntu-latest' }}
+        if: ${{ matrix.beam_versions.otp == env.RUN_DIALYZER_ON_OTP_RELEASE && matrix.os == 'ubuntu-latest' }}
         with:
           path: _build/*/rebar3_*_plt
           key: dialyzer-plt-cache-${{ steps.install-erlang.outputs.otp-version }}-${{ runner.os }}-${{ hashFiles('rebar.config*') }}-v1
@@ -44,12 +51,17 @@ jobs:
       - name: Coveralls
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          COVERALLS_FLAG_NAME: Erlang/OTP ${{ matrix.otp_version }} on ${{ matrix.os }}
+          COVERALLS_FLAG_NAME: Erlang/OTP ${{ matrix.beam_versions.otp }} on ${{ matrix.os }}
         run: rebar3 as test coveralls send
 
       - name: Dialyzer
-        if: ${{ matrix.otp_version == env.RUN_DIALYZER_ON_OTP_RELEASE && matrix.os == 'ubuntu-latest' }}
+        if: ${{ matrix.beam_versions.otp == env.RUN_DIALYZER_ON_OTP_RELEASE && matrix.os == 'ubuntu-latest' }}
         run: rebar3 clean && rebar3 as test dialyzer
+
+      - name: Get Elixir dependencies
+        run: mix deps.get
+      - name: ExUnit
+        run: mix test
 
   Finish:
     name: Finishing
--- a/mix.exs
+++ b/mix.exs
@@ -14,7 +14,7 @@ defmodule Khepri.MixProject do
   defp deps() do
     [
       # Dependency pinning must be updated in rebar.config too.
-      {:ra, "2.0.4"}
+      {:ra, "2.0.9"}
     ]
   end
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -6,12 +6,18 @@ on:
...
     strategy:
       matrix:
-        otp_version: [23, 24, '25.0-rc3']
+        beam_versions:
+          - otp: 23
+            elixir: '1.11.4'
+          - otp: 24
+            elixir: '1.13.4'
+          # - otp: '25.0-rc3'
+          #   elixir: '1.13.4'
```
the-mikedavis, line 20: The installation when using this combination errors out. Ideally, this section in the matrix would skip all Elixir steps and only run the Erlang parts.

mix.exs
```diff
@@ -11,10 +11,14 @@ defmodule Khepri.MixProject do
...
 
+  def application do
+    [extra_applications: [:elixir]]
+  end
+
   defp deps() do
     [
       # Dependency pinning must be updated in rebar.config too.
-      {:ra, "2.0.4"}
+      {:ra, "2.0.9"}
```
the-mikedavis, line 21: It looks like this fell behind a bit. In Elixir it's very common to use the syntax from [`Version`](https://hexdocs.pm/elixir/Version.html) to enforce dependency versions. Typically these follow semantic versioning rules so you might use

```elixir
{:ra, "~> 2.0"}
```

to enforce "ra at >= 2.0.0 and < 3.0.0" or

```elixir
{:ra, "~> 2.0.0"}
```

for "ra at >= 2.0.0 and < 2.1.0". I think rebar handles this version syntax these days as well.

If you want to be exact,

```elixir
{:ra, "== 2.0.9"}
```

would be the typical way to say it.

dumbbell: Indeed, despite the "Dependency pinning must be updated in mix.exs too." I wrote in `rebar.config`, I forgot to update `mix.exs`...

Perhaps our `mix.exs` can be improved to read the information it needs from `rebar.config` and `src/khepri.app.src`. I will try to implement that because this would solve the problem of keeping them in sync.

Thanks, I will use `"== 2.0.9"`. The reason I don't want to loosen the version pinning is that I don't follow the development of Ra closely. There are no release notes and we are bad at release engineering in the RabbitMQ team (I try to improve that in Khepri), meaning I'm not 100% sure Ra follows semantic versioning.

dumbbell (commented):
Thank you for this experiment!

I admit I don't want to involve Elixir in the build process of Khepri. I find it's a nightmare to have to support both in RabbitMQ, because there is always a case where Elixir was built against a newer version of Erlang compared to the one being used and the build fails. I don't want to reproduce that in Khepri.

For this Elixir-specific API, I feel I need to collect more details to make up my mind. I would love to provide directly in Khepri the tools to make Elixir developer's life easy with Khepri. This sigil support looks to be one of them. But if I understand correctly, either we only provide the function-based runtime one in an Erlang module, or we provide both the compile-time and runtime versions but we need an Elixir module for the former.

dumbbell:
You say:
> (8e9483a has notes on the difference between lowercase and uppercase.)

> As a side-note, lowercase sigil names allow string interpolation while uppercase do not, which is why `sigil_p/2` must work at runtime rather than compile-time.

And the documentation confirms:
> If the sigil letter is in uppercase, no interpolation is allowed in the sigil, otherwise its contents may be dynamic.

I suppose this is implemented as:
1. if the sigil letter is in lowercase, do string interpolation
2. pass the string to the sigil implementation

So I understand that the lowercase version must be a runtime-evaluated sigil. For the uppercase copy, I guess that it could be either a compile-time sigil macro or a runtime function-based sigil, is that correct?

the-mikedavis:
Yep, spot on! The trouble with working with the lowercase version at compile-time is that the string interpolation isn't evaluated until runtime. If you inspect the AST:

```elixir
defmodule Khepri do
  defmacro sigil_p(path, _opts) do
    IO.inspect(path, label: "ast")
    quote do
      :khepri_path.from_binary(unquote(path))
    end
  end

  defmacro sigil_P({:<<>>, _, [path]} = input, _opts) do
    IO.inspect(input, label: "ast")
    path
    |> :khepri_path.from_binary()
    |> Macro.escape()
  end
end

iex> import Khepri
iex> kind = "wood"
iex> ~p"/stock/#{kind}/:oak"
ast: {:<<>>, [line: 6],
 [
   "/stock/",
   {:"::", [line: 6],
    [
      {{:., [line: 6], [Kernel, :to_string]}, [line: 6],
       [{:kind, [if_undefined: :apply, line: 6], nil}]},
      {:binary, [if_undefined: :apply, line: 6], nil}
    ]},
   "/:oak"
 ]}
["stock", "wood", :oak]
iex> ~P"/stock/#{kind}/:oak"
ast: {:<<>>, [line: 6], ["/stock/\#{kind}/:oak"]}
["stock", "\#{kind}", :oak]
```

We don't have enough information at compile-time to transform the string if there's interpolation in it.

And yep, you could also implement `sigil_P/2` as a regular function. The benefit to doing it at compile-time is that any failures to compile the path will cause a compilation error rather than a runtime error.

the-mikedavis:
> But if I understand correctly, either we only provide the function-based runtime one in an Erlang module, or we provide both the compile-time and runtime versions but we need an Elixir module for the former.

We could provide both `sigil_p/2` and `sigil_P/2` as runtime functions and have them implemented in Erlang. That would also help the documentation: I'm not sure it's possible to combine Erlang and Elixir docs in a setup like this so implementing everything in Erlang seems like the best approach.

the-mikedavis:
> I admit I don't want to involve Elixir in the build process of Khepri. I find it's a nightmare to have to support both in RabbitMQ, because there is always a case where Elixir was built against a newer version of Erlang compared to the one being used and the build fails. I don't want to reproduce that in Khepri.

I agree, it's definitely more of a headache than it's worth 👍

Let me close this out but please feel free to continue the discussion / ask questions 😀

PR #100, opened 2022-05-04
mix.exs: Read application details from Rebar files

dumbbell:
The information we need is:
* the application name
* its description
* its version
* its list of dependencies and their version pinning

Everything is available from `src/khepri.app.src` and `rebar.config`.

We do a very basic read of those files. In other words, we don't bother with Rebar profiles, `rebar.config.script` interpretation and so on. We expect a very straightforward list of dependencies, but that should be fine for our needs.

This should solve the problem of keeping `mix.exs` in sync. Indeed, this was triggered by the fact that Ra's version pinning in `mix.exs` was out-of-date for some time now.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -2,19 +2,48 @@ defmodule Khepri.MixProject do
   use Mix.Project
 
   def project do
+    # To avoid duplication, we query the app file to learn the application
+    # name, description and version.
+    {:ok, [app]} = :file.consult("src/khepri.app.src")
+    {:application, app_name, props} = app
+
+    description = to_string(Keyword.get(props, :description))
+    version = to_string(Keyword.get(props, :vsn))
+
     [
-      app: :khepri,
-      description: "Tree-like replicated on-disk database library",
-      version: "0.3.0",
+      app: app_name,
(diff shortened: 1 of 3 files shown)


dumbbell:
@the-mikedavis: What do you think of this approach?

the-mikedavis (approved):
This looks perfect!

It could be more Elixir-y by replacing `proplists` calls with `Keyword` functions or [`Lists.keyfind/4`](https://hexdocs.pm/elixir/List.html#keyfind/4) where appropriate and `:erlang.list_to_binary/1` with [`to_string/1`](https://hexdocs.pm/elixir/Kernel.html#to_string/1) but on the other hand, given that Khepri is an Erlang library, I think it's better to use the Erlang stdlib rather than Elixir so it's more approachable to pure-Erlang devs.

dumbbell:
I went and used the Elixir modules & functions. I didn't know what to use and couldn't find them with a quick search. Thank you for pointing me to them!

This file is mostly for Elixir developers anyway. Erlang developers will look at the Rebar configuration mostly I suppose.

PR #124, opened 2022-08-03
Setup bazel build files

the-mikedavis:
_This is a draft pending changes in `rules_erlang`: https://github.com/rabbitmq/rules_erlang/pull/61_

This sets up bazel build files mostly influenced by osiris and ra.

There was an adjustment to make in the test cases since the stacktrace filenames differ between rebar3 and bazel.


the-mikedavis:
Closing this since we should be able to keep the bazel build definition cleanly in the server repo. It nice to have just rebar for building khepri - bazel introduces some duplication of work when you need to update the version or dependencies. We can avoid that by having the build definition in the server repo only.

PR #125, opened 2022-08-08
Handle `nodedown` on process_query_response

dcorbacho:
Leader needs to be updated

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -492,16 +492,18 @@ process_sync_command(StoreId, Command, Options) ->
             Ret;
         {timeout, _} = TimedOut ->
             {error, TimedOut};
-        {error, noproc}
-          when LeaderId =/= undefined andalso ?HAS_TIME_LEFT(Timeout) ->
+        {error, Reason}
+          when LeaderId =/= undefined andalso ?HAS_TIME_LEFT(Timeout) andalso
+               (Reason == noproc orelse Reason == nodedown) ->
             %% The cached leader is no more. We simply clear the cache
             %% entry and retry.
             khepri_cluster:clear_cached_leader(StoreId),
             NewTimeout = khepri_utils:end_timeout_window(Timeout, T0),
             Options1 = Options#{timeout => NewTimeout},
             process_sync_command(StoreId, Command, Options1);
-        {error, noproc} = Error
-          when LeaderId =:= undefined andalso ?HAS_TIME_LEFT(Timeout) ->
+        {error, Reason} = Error
+          when LeaderId =:= undefined andalso ?HAS_TIME_LEFT(Timeout) andalso
+               (Reason == noproc orelse Reason == nodedown) ->
             case khepri_utils:is_ra_server_alive(RaServer) of
                 true ->
                     %% The follower doesn't know about the new leader yet.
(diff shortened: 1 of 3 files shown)


dumbbell:
Coud you please 1) add a testcase and 2) prepend "khepri_machine: " to the commit message to indicate the context/scope?

src/khepri_machine.erl
```diff
@@ -667,16 +667,18 @@ process_query_response(
     {error, TimedOut};
 process_query_response(
   StoreId, _RaServer, true = _IsLeader, QueryFun, QueryType, Timeout,
-  {error, noproc})
-  when QueryType =/= local andalso ?HAS_TIME_LEFT(Timeout) ->
+  {error, Reason})
+  when QueryType =/= local andalso ?HAS_TIME_LEFT(Timeout) andalso
+       (Reason == noproc orelse Reason == nodedown) ->
```
dumbbell, line 672: The same should be applied to `process_sync_command/3` with an associated testcase.

test/cluster_SUITE.erl
```diff
@@ -1071,10 +1151,10 @@ start_n_nodes(NamePrefix, Count) ->
                  ct:pal("- ~s", [Name]),
                  start_erlang_node(Name)
              end || I <- lists:seq(1, Count)],
-    ct:pal("Started nodes: ~p", [Nodes]),
+    ct:pal("Started nodes: ~p", [[N || {_P, N} <- Nodes]]),
```
dumbbell, line 1154: `_P` and `N` are swapped compared to what `start_erlang_node/1` returns.

Also, could you please use the same variable name as elsewhere, so `_Peer` and `Node`?

test/cluster_SUITE.erl
```diff
@@ -1083,15 +1163,20 @@ start_n_nodes(NamePrefix, Count) ->
...
-    Node.
+    {Node, Peer}.
+stop_erlang_node(_Node, Peer) ->
+    ok = peer:stop(Peer).
 -else.
 start_erlang_node(Name) ->
     Name1 = list_to_atom(Name),
     Options = [{monitor_master, true}],
     {ok, Node} = ct_slave:start(Name1, Options),
     Node.
```
dumbbell, line 1176: This copy of `start_erlang_node/1` for Erlang<25 does not return the same thing as the copy for Erlang>=25.

It could return `{Node, Node}` where the second argument is an opaque value required by the stop function.

test/cluster_SUITE.erl
```diff
@@ -1083,15 +1163,20 @@ start_n_nodes(NamePrefix, Count) ->
 -if(?OTP_RELEASE >= 25).
 start_erlang_node(Name) ->
     Name1 = list_to_atom(Name),
-    {ok, _, Node} = peer:start(#{name => Name1,
+    {ok, Peer, Node} = peer:start(#{name => Name1,
                                  wait_boot => infinity}),
```
dumbbell, line 1167: The indentation of this line needs to be adjusted with the change of the line above.

dumbbell (changes requested):

test/cluster_SUITE.erl
```diff
@@ -662,6 +668,157 @@ can_restart_nodes_in_a_three_node_cluster(Config) ->
...
+    Peer = proplists:get_value(StoppedLeaderNode1, PeerPerNode),
+
+    ct:pal(
+      "Stop database on leader node ~s (quorum is maintained)",
+      [StoppedLeaderNode1]),
+    ?assertEqual(ok, stop_erlang_node(StoppedLeaderNode1, Peer)),
+
+    lists:foreach(
+      fun(Node) ->
+              ct:pal("- khepri:get() from node ~s", [Node]),
```
dumbbell, line 738: This logged message doesn't match the line below :-)

dumbbell (changes requested):
Could you please rebase onto the latest `main`. There was a fix in GitHub Actions.

dumbbell:
Thank you! I will add a local change I have to improve code coverage with multiple Erlang nodes and I will merge your branch.

dumbbell (approved):

PR #126, opened 2022-08-08
khepri: Return errors in exists(), has_data() and has_sproc()

dumbbell:
Before this change, all errors were considered the same and `false` was returned. Now, only the `{error, {node_not_found, _}}` error is interpreted and `false` is returned. All other errors are returned as is.

This helps callers who need to know if the given path is incorrect or if the Ra cluster is degraded.


the-mikedavis (approved):
There's a tiny typo in the commit message:

```diff
-This helps callers who need to know if the given path is incorrect or of
+This helps callers who need to know if the given path is incorrect or if
 the Ra cluster is degraded.
```

but otherwise I think this looks good - it's much more user-friendly that `khepri:exists("*")` returns an error rather than just `false` 👍

PR #129, opened 2022-08-08
Add an erlang_ls configuration file

the-mikedavis:
This file is picked up by erlang_ls when running through an editor.
Without it, erlang_ls is functional but it will mistakenly jump to
`_build/default/lib/khepri/src/` when jumping to definitions or
references within the source.


the-mikedavis:
Actually erlang_ls is working for me without a local `erlang_ls.config` file here. I had a stray `~/.config/erlang_ls/erlang_ls.config` file that was confusing erlang_ls about where to find files. Sorry for the noise!

PR #130, opened 2022-08-15
khepri_fun: Use locks to avoid concurrent calls to code:get_object_code/1 for the same fun

dcorbacho:
Parallel calls to a function that is not yet in the cache will call simultaneously `code:get_object_code/1`. This call then becomes very slow and then the cache update also happens in parallel, with the same value. 
Ensuring that only one process tries to process the function and update the cache yields much faster parallel execution times, while the other process then also hit the new cache value and not the processing function path.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -323,10 +323,45 @@ to_standalone_fun1(Fun, Options) ->
       Fun :: fun(),
       State :: #state{},
       StandaloneFun :: standalone_fun().
+%% @doc Extracts the given anonymous function
+%%
+%% Parallel calls to handle a function that is not yet in the cache will
+%% simultaneously call {@link code:get_object_code/1}. This server call then
+%% becomes very slow, and later all the processes still have to update the
+%% cache - with exactly the same value. Ensuring that only one process tries to
+%% process the function and update the cache yields much faster parallel
+%% execution times, while the other process then also hit the new cache value
+%% and not the processing function path.
+%%
 %% @private
 %% @hidden
 
 to_standalone_fun2(Fun, State) ->
+    case to_cached_standalone_fun(Fun, State) of
+        undefined ->
+            Lock = {{khepri_fun, Fun}, self()},
+            global:set_lock(Lock, [node()]),
+            try
+                case to_cached_standalone_fun(Fun, State) of
+                    undefined             -> to_standalone_fun3(Fun, State);
+                    StandaloneFunAndState -> StandaloneFunAndState
+                end
+            after
+                global:del_lock(Lock, [node()])
+            end;
+        StandaloneFunAndState ->
+            StandaloneFunAndState
+    end.
+
+-spec to_cached_standalone_fun(Fun, State) ->
+    {StandaloneFun, State} | undefined when
+      Fun :: fun(),
+      State :: #state{},
+      StandaloneFun :: standalone_fun().
+%% @private
+%% @hidden
+
+to_cached_standalone_fun(Fun, State) ->
     case get_cached_standalone_fun(State) of
         #standalone_fun{} = StandaloneFunWithoutEnv ->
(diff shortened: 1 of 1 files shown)


lukebakken (commented):
Is this a good place to use `try ... after`?
```
to_standalone_fun2_with_retry(Fun, State) ->
    case to_standalone_fun2(Fun, State) of
        undefined ->
            Lock = {{khepri_fun, Fun}, self()},
            try
                global:set_lock(Lock, [node()]),
                case to_standalone_fun2(Fun, State) of
                    undefined ->
                        to_standalone_fun3(Fun, State);
                    Other ->
                        Other
                end
            after
                global:del_lock(Lock, [node()])
            end;
        Other ->
            Other
    end.
```

the-mikedavis (approved):

lukebakken (approved):
Nice fix @dcorbacho!

src/khepri_fun.erl
```diff
@@ -317,9 +317,9 @@ to_standalone_fun1(Fun, Options) ->
     State0 = #state{fun_info = Info,
                     all_calls = #{{Module, Name, Arity} => true},
                     options = Options},
-    to_standalone_fun2(Fun, State0).
+    to_standalone_fun2_with_retry(Fun, State0).
 
--spec to_standalone_fun2(Fun, State) -> {StandaloneFun, State} when
+-spec to_standalone_fun2(Fun, State) -> {StandaloneFun, State} | undefined when
```
dumbbell, line 322: Could you please rename this function to `to_standalone_fun2_locked()` and call the new function `to_standalone_fun2()`?

This is a common convention so we understand that `to_standalone_fun2()` and `to_standalone_fun2_locked()` are the same thing, and `to_standalone_fun2()` is responsible for wrapping the call to `to_standalone_fun2_locked()` in a lock.

src/khepri_fun.erl
```diff
@@ -337,7 +337,26 @@ to_standalone_fun2(Fun, State) ->
         fun_kept ->
             {Fun, State};
         undefined ->
-            to_standalone_fun3(Fun, State)
+            undefined
+    end.
+
+to_standalone_fun2_with_retry(Fun, State) ->
```
dumbbell, line 343: Could you please move this function before `to_standalone_fun2_locked()` (instead of after)?

src/khepri_fun.erl
```diff
@@ -337,7 +337,26 @@ to_standalone_fun2(Fun, State) ->
...
+    case to_standalone_fun2(Fun, State) of
+        undefined ->
+            Lock = {{khepri_fun, Fun}, self()},
+            try
+                global:set_lock(Lock, [node()]),
```
dumbbell, line 348: The lock must be acquired before the try/catch because if this call raises an exception, `global:del_lock()` will be called despite no acquired lock.

src/khepri_fun.erl
```diff
@@ -337,7 +337,26 @@ to_standalone_fun2(Fun, State) ->
...
+                    undefined ->
+                        to_standalone_fun3(Fun, State);
+                    Other ->
+                        Other
+                end
+            after
+                global:del_lock(Lock, [node()])
+            end;
+        Other ->
+            Other
```
dumbbell, line 359: Could you please name the variable `StandaloneFunAndState` for instance? I feel like `Other` could be any random garbage returned by the function, but we always expect a success return value.

dumbbell (changes requested):
Could you please add a comment in the code to explain why we need a lock?

I think you can squash all commits into one and repeat the same explanation in the commit message.

dumbbell (approved):

dumbbell:
I changed my mind w.r.t to the `_locked` suffix as I didn't read the patch correctly first, sorry.

I made a few cosmetic changes and will merge your pull request once CI is done.

Thank you!

PR #132, opened 2022-08-17
Remove OTP23 from CI workflows

the-mikedavis:
From a Slack discussion: OTP23 is EOL for the server so we should be able to drop support here. There aren't any failures on OTP23 but we may be able to remove some compatibility code from `khepri_fun`. (I'm not sure of any off the top of my head but we should be able to figure it out with coverage reports between runs with OTP23-25)


dumbbell (approved):
Thank you!

PR #134, opened 2022-08-22
khepri_fun: Handle FunIndex variant of call_fun2/3

the-mikedavis:
call_fun2/3 has three possible tags for the second element in the
instruction tuple:

    ## @spec call_fun2 Tag Arity Func
    ## @doc Calls the fun Func with arity Arity. Assume arguments in registers x(0)
    ##      to x(Arity-1). Tag can be one of:
    ##
    ##      * FunIndex      - `Func` is always a local fun identified by `FunIndex`
    ##      * {atom,safe}   - `Func` is known to be a fun of correct arity.
    ##      * {atom,unsafe} - Nothing is known about `Func`.
    178: call_fun2/3

We already handle the {atom,safe} variant (#86). By nesting some higher
order functions and using a local function as one of the called
functions, we can produce the FunIndex variant.

When translated correctly, this variant has a {f,Label} shape
where the Label is known to be the beginning of the called local
function's scope.

    {call_fun2,
     {f,821},
     3,
     {tr,{x,3}, {t_fun,3,{'-projection_fun_for_sets/1-fun-1-',4},any}}}.

But beam_disasm gives the shape:

    {call_fun2,
     {u,57},
     3,
     {tr, {x,3}, {{t_fun,any,any,any},0,18446744073709551615}}.

The type-tagged register must be decoded to discard the surrounding
information similar to other instructions that use type-tagged
registers.

The {u,?} register doesn't point to a recognizable label index so
we can't determine the correct function index to use. However we
know that if the compiler emitted the FunIndex variant of
call_fun2/3, it must know that a local function of the correct arity
is being called. So we can replace the {u,?} tag with {atom,safe}
which signals that the function is known to be the correct arity.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -1114,6 +1114,24 @@ pass1_process_instructions(
     Instruction = {call_fun2, {atom, safe}, Arity, {tr, FunReg, Type}},
     pass1_process_instructions([Instruction | Rest], State, Result);
 pass1_process_instructions(
+  [{call_fun2,
+    {u, _},
+    Arity,
+    {tr, FunReg, {{t_fun, _Arity, _Domain, _Range} = Type, _, _}}} | Rest],
+  State,
+  Result) ->
+    %% `beam_disasm' did not decode this instruction correctly. The
+    %% type in the type-tagged record is wrapped with extra information
+    %% we discard. We also need to patch it to remove the referenced
+    %% FunIndex. We don't have enough information to lookup the local
+    %% function referenced by the live register but if the compiler knows
+    %% that there is an exact local function used in this call that it
+    %% must have the correct arity, so we can demote the function info
+    %% part of this instruction to `{atom, safe}'. That tag asserts that
+    %% the function is known to be the correct arity.
+    Instruction = {call_fun2, {atom, safe}, Arity, {tr, FunReg, Type}},
+    pass1_process_instructions([Instruction | Rest], State, Result);
+pass1_process_instructions(
   [{test, BsGetSomething,
     Fail, [Ctx, Live, Size, Unit, {field_flags, FF} = FieldFlags0, Dst]}
    | Rest],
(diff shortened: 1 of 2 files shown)


dumbbell:
Are we sure about that?

This `FunIndex` must be an index in one of the tables inside the beam file. I will look into this when I'm back from holidays.

dumbbell:
I looked more carefully that `call_fun2` instruction. I believe your analysis is correct.

In your testcase, the instruction looks like this:
```erlang
{call_fun2, {u,2}, 3, {tr, {x,3}, {t_fun, any, any, any}}}
```

The `{x,3}` register is enough to jump to the correct function.

`{u,2}` is indeed an offset into the "Lambda table", named `"FunT"` in the Beam file. I decoded that table from `tx_funs.beam`, here is the 3rd entry (index 2 starting from 0):
```erlang
#lambda{
  function = '-projection_fun_for_sets/1-fun-1-',
  num_free = 1,
  arity = 4,
  label = 893,
  index = 2,
  old_uniq = 125568841}
```

We could convert the label of 893 to the new location in the generated module. I suppose the generated code would be slightly optimized.

Note that the decoded arity of 4 doesn't match the arity of the actual function (3). That's because the function takes a variable from its environment. Therefore it has 4 inputs: 3 from the arguments, 1 from the environment.

We also have enough details to recreate the tagged type by the way:
```erlang
{t_fun, 3, {'-projection_fun_for_sets/1-fun-1-', 4}, any}
```

I'm undecided if it's worth it though. What do you think, @the-mikedavis?

the-mikedavis:
If we have enough information to decode the instruction fully, I would prefer to use it. At least that way the eventual PRs upstream to fix beam_disasm can be more complete, eh?

dumbbell:
Ok, I will prepare a follow-up branch with what I have. Thanks!

dumbbell:
@the-mikedavis: I opened #149 with my patch.

PR #135, opened 2022-08-22
Add projections

the-mikedavis:
Projections are a system for creating replicated ETS caches available on all members of a Khepri cluster. A new function `khepri_projection:new/3` creates a projection resource:

```erl
ProjectionName = wood_stocks,
ProjectionFun = fun([stock, wood, Kind], Stock) -> {Kind, Stock} end,
Options = #{type => set, read_concurrency => true},
Projection = khepri_projection:new(ProjectionName, ProjectionFun, Options).
```

The projection resource may then be registered to a pattern of nodes with `khepri:register_projection/4`:

```erl
StoreId = stock,
PathPattern = "/:stock/:wood/*",
RegisterOptions = #{},
khepri:register_projection(
  StoreId, PathPattern, Projection, RegisterOptions).
```

Once registered, the projection's ETS table is created and then filled with any existing records matching the pattern. Paths and data are passed through the `ProjectionFun` to create records which are then stored in the table. Any future changes to records which match the `PathPattern` are also applied to the projection table.

The projection table is a named ETS table which may be queried directly with `ets`:

```erl
khepri:put(StoreId, "/:stock/:wood/oak", 100),
ets:lookup_element(ProjectionName, <<"oak">>, 2).
%%=> 100
```

Projections should be used to maximize query throughput and/or minimize query latency at the cost of consistency and memory consumption. Projections have the same consistency guarantees as queries tried with the `#{favor => low_latency}` option. Data stored in the projection table is at least partially duplicated between the store and the projection table, so increased memory consumption is expected for projections with path patterns matching many nodes.


src/khepri_projection.erl
```diff
@@ -0,0 +1,296 @@
...
+%% `Payload' is `undefined' if there is no payload for the node.
+%%
+%% This function is compiled the same way as a transaction function: all
+%% side-effects are not allowed.
+
+-type extended_projection_fun() ::
+fun((Table :: ets:tid(),
+     Path :: khepri_path:native_path(),
+     OldPayload :: term() | undefined,
+     NewPayload :: term() | undefined) -> any()).
```
dumbbell, line 77: I find it a bit confusing that the payload goes through several forms between the state machine and the final call to the projection function.

What do you think of using a node properties map (see `khepri_machine:gather_node_props()`) to transport the old and new payloads all to the way to the call of the projection function? This way, you can distinguish between a stored `undefined` atom and the absence of data payload.

src/khepri_projection.erl
```diff
@@ -0,0 +1,296 @@
...
+    if
+        is_function(Fun, 2) ->
+            FunOptions = #{ensure_instruction_is_permitted =>
+                           fun khepri_tx_adv:ensure_instruction_is_permitted/1,
+                           should_process_function =>
+                           fun khepri_tx_adv:should_process_function/4,
+                           is_standalone_fun_still_needed =>
+                           fun(_Params) -> true end},
+            StandaloneFun = khepri_fun:to_standalone_fun(Fun, FunOptions),
+            {simple, StandaloneFun};
```
dumbbell, line 147: You could still query the arity of the extracted function later. I wonder if it's necessary to wrap the function in a `{simpl | extended, StandaloneFun}` tuple. Perhaps it's more future proof.

src/khepri_projection.erl
```diff
@@ -0,0 +1,296 @@
...
+%% @hidden
+%% Applies the projection function against the entry from the tree to return
+%% a projected record. This projected record is then applied to the ETS table.
+
+trigger(
+  #?MODULE{name = Name, project_fun = ProjectFun},
+  Path, OldPayload0, NewPayload0) ->
+    Table = ets:whereis(Name),
+    OldPayload = ?PAYLOAD_OR(OldPayload0, undefined),
+    NewPayload = ?PAYLOAD_OR(NewPayload0, undefined),
```
dumbbell, line 234: See my comment earlier. In the rest of the library, we use the node properties map as the structure to carry the node details around, even if it's only the payload.

src/khepri_projection.erl
```diff
@@ -0,0 +1,296 @@
...
+            ok
+    end.
+
+-spec trigger_simple_projection(
+        Table, StandaloneFun, Path, Payload, Payload) ->
+    Ret when
+      Table :: ets:tid(),
+      StandaloneFun :: khepri_fun:standalone_fun(),
+      Path :: khepri_path:native_path(),
+      Payload :: term() | ?NO_PAYLOAD,
```
dumbbell, line 270: As far as I understand, `?NO_PAYLOAD` is never passed as it is already converted to `undefined`.

src/khepri_tx_adv.erl
```diff
@@ -42,9 +42,12 @@
          clear_payload/1, clear_payload/2,
          clear_many_payloads/1, clear_many_payloads/2]).
 
-%% For internal user only.
+%% For internal use only.
```
dumbbell, line 45: Thanks! I committed this fix in `main` to not pollute this patch.

dumbbell (changes requested):
I'm so glad about this addition! This is really cool :-)

I submitted several comments. I could split my feedback into two topics:
* Some documentation comments
* The way payloads are passed from the state machine to the projection function

About the payloads, I suggest we do like the rest of the library: call `khepri_machine:gather_node_props()` to create a node properties map and pass this common format to `khepri_projection:trigger()`, then decide in that function how to call the projection function.

I think we could pass the entire node properties maps (old and new) to the extended projection function, and reduce payloads to bare data for the simple one (possibly `undefined` for the old data in case a payload is added). The simple function will never be called when the new payload has no data.

What do you think?

src/khepri.erl
```diff
@@ -2789,6 +2905,7 @@ info(StoreId) ->
 %% @doc Lists the content of specified store on <em>stdout</em>.
 %%
 %% @param StoreId the name of the Khepri store.
+%% @param Options query options.
```
the-mikedavis, line 2908: I'll push this one to master separately as well since it's unrelated to the change

rebar.lock
```diff
@@ -1,17 +1,17 @@
 {"1.2.0",
 [{<<"aten">>,{pkg,<<"aten">>,<<"0.5.8">>},1},
  {<<"gen_batch_server">>,{pkg,<<"gen_batch_server">>,<<"0.8.8">>},1},
- {<<"ra">>,{pkg,<<"ra">>,<<"2.1.0">>},0},
+ {<<"ra">>,{pkg,<<"ra">>,<<"2.4.0">>},0},
```
the-mikedavis, line 4: 2.4.1 is released now as well (it bumps seshat). I can bump from 2.4.0->2.4.1 after this is merged since the commit that bumps to 2.4.0 has details on what in 2.4.0 is necessary for this feature

the-mikedavis:
Thanks for the feedback @dumbbell!

I like the idea of passing `khepri:node_props()` around rather than payloads :+1:. I switched to that and now `khepri_projection:extended_projection_fun()`s take full node-props maps too.

For the `{simple | extended, Fun}` part: originally I thought we could avoid using standalone funs for simple projections (we can't because the funs are persisted to the log) and that tag is a relic from that. I switched to just holding on to a standalone fun and checking the arity.

rebar.lock
dumbbell: Ok

src/khepri_projection.erl
```diff
@@ -0,0 +1,282 @@
...
+  Table, StandaloneFun, Path, OldProps, NewProps) ->
+    case {OldProps, NewProps} of
+        {#{data := OldPayload}, #{data := NewPayload}} ->
+            OldRecord = khepri_fun:exec(StandaloneFun, [Path, OldPayload]),
+            case khepri_fun:exec(StandaloneFun, [Path, NewPayload]) of
+                OldRecord ->
+                    ok;
+                NewRecord ->
+                    ets:delete_object(Table, OldRecord),
+                    ets:insert(Table, NewRecord)
```
dumbbell, line 271: There is tiny window during which the record is missing from the ETS table. This is nearly impossible that a read happens during this window, therefore it will happen often :-)

For simple projection functions, I wonder if we should have stricter conditions on the ETS table configuration and use. Like, bags are disallowed and the key must remain the same for the same input path? This way, we don't have to remove the old record. What do you think?

src/khepri_projection.erl
```diff
@@ -0,0 +1,282 @@
...
+%% `OldPayload' or `NewPayload' are empty maps if there is no tree node. For
+%% example, a newly created node will have an empty map for `OldPayload' and
+%% a {@link khepri:node_props()} map with values for `NewPayload'.
+%%
+%% This function is compiled like a transaction function except that calls
+%% to the {@link ets} module are allowed.
+
+-type projection_fun() :: simple_projection_fun() | extended_projection_fun().
+%% A function that formats an entry in the tree into a record to be stored
+%% in a projection.
```
dumbbell, line 89: If I understand correctly, projection functions run in the context of the Ra server process.

We should document that and specify that the function should be simple and fast if possible.

src/khepri_projection.erl
```diff
@@ -0,0 +1,282 @@
...
+-type extended_projection_fun() ::
+fun((Table :: ets:tid(),
+     Path :: khepri_path:native_path(),
+     OldPayload :: khepri:node_props(),
+     NewPayload :: khepri:node_props()) -> any()).
+%% An extended projection function.
+%%
+%% In some cases, a node in the store might correspond to many objects in a
+%% projection table. Extended projection functions are allowed to call ETS
+%% functions directly in order to build the projection table.
```
dumbbell, line 78: We could specify that the return value of the extended projection function is ignored.

src/khepri_projection.erl
```diff
@@ -0,0 +1,282 @@
...
+                                "  New props:~n"
+                                "    ~p~n"
+                                "  Crash:~n"
+                                "    ~ts",
+                                [Name, Path, OldProps, NewProps,
+                                khepri_utils:format_exception(
+                                  Class, Reason, Stacktrace, #{column => 4})]),
+            ?LOG_ERROR(Msg, []),
+            ok
+    end.
```
dumbbell, line 248: I think the try/catch block should wrap the various calls to `khepri_fun:exec()` only. Otherwise it will mask crashes in Khepri itself.

dumbbell (changes requested):
Thanks for the changes, it looks good to me. I have a few more tiny comments on the documentation and one question for the simple projection function.

src/khepri_projection.erl
```diff
@@ -0,0 +1,282 @@
...
+
+ %% For internal use only
+-export([init/1, trigger/4]).
+
+-export_type([projection/0,
+              projection_fun/0,
+              options/0]).
+
+-record(?MODULE,
+        {%% The name of the projection. This atom should be unique to a node
```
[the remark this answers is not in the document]
the-mikedavis: For this one I meant "Erlang node" so I clarified that

src/khepri_machine.erl
```diff
@@ -1766,11 +1901,86 @@ delete_matching_nodes_cb(_, {interrupted, _, _}, _Options, Result) ->
...
+                               InitialRoot
+                       end,
+    case does_path_match(Path, Pattern, [], PathMatchingRoot) of
+        true ->
+            FindOptions = #{props_to_return => [payload_version,
+                                                child_list_version,
+                                                child_list_length,
+                                                child_names,
+                                                payload,
+                                                has_payload],
```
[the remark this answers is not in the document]
the-mikedavis: I put these two into a macro so they're reused 👍

src/khepri_projection.erl
dumbbell: Yes, I agree, documenting this constraint is good enough.

dumbbell (approved):

src/khepri_projection.erl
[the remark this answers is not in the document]
dumbbell: Oops, yes, my mistake.

PR #136, opened 2022-08-25
khepri_machine: add 'configure' command

dcorbacho:
Allows to change snapshot_interval during runtime.
Accepts a map of options so other parameters could be changed later on

### DIFF
--- a/src/internal.hrl
+++ b/src/internal.hrl
@@ -71,3 +71,5 @@
                     event_filter :: khepri_evf:event_filter(),
                     sproc :: khepri_fun:standalone_fun(),
                     props = #{} :: map()}).
+
+-record(configure, {params = #{} :: map()}).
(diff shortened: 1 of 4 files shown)


dumbbell:
Can we please revert this?
* I would like us to discuss this first.
* There is a mismerge somewhere because it reintroduces code that @the-mikedavis removed.
* There is no associated testsuite.

dcorbacho:
I'll revert it.

You can see the thought behind this here: https://github.com/rabbitmq/rabbitmq-server/issues/5515#issuecomment-1227513349
I just saw the `OTP24` thing - I don't know how it went through. 
I forgot to add the testcase to the commit, it's a new suite. My bad.

dumbbell:
Thank you for the revert.

We started to talk about ideas (such as not accounting for `put` of new entries for snapshots) to let Khepri balance snapshot requests automatically to improve performance of large imports. I think we should pursue this research instead of introducing a command. But as said in my previous comment, we should discuss it so we all agree on the path :-)

PR #150, opened 2022-10-13
Support stored procedures as transaction functions

### LINKED ISSUE #141: Accept a stored procedure as a transaction function
Currently, transaction functions must be anonymous functions with an arity of 0:
```erlang
khepri:transaction(StoreId, fun() -> ... end).
```

However, when the same anonymous function is used over and over again, this is not optimal:
1. Khepri will attempt a function extraction (but will hit the cache to save extraction and compile time).
2. A large Ra command will be built (with the entire function's compiled binary embedded) and appended to Ra's log for every calls, even though the function remains the same (including its environment).

It would be nice to be able to store the anonymous function once as a stored procedure, then call this stored procedure as the transaction function:
```erlang
StoredProcPath = "/:path/:to/:stored/:proc",
khepri:put(StoreId, StoredProcPath, fun(Arg1, Arg2) -> ... end),
khepri:transaction(StoreId, StoredProcPath, [my_arg1, my_arg2]).
```

One difference is that anonymous functions passed to `khepri:transaction/2` are expected to take no arguments. They take "arguments" as part of their environment. This is not enough if we use stored procedure. Therefore, the `khepri:transaction/2` API will have to evolve to also take a list of arguments to pass to the stored procedure.

dumbbell:
This patch brings two improvements:

1.  A **transaction function can take arguments**. Arguments are passed as a list of terms to `khepri:transaction()`.

    Here is an example:

    ```erlang
    generic_tx(Wood, Increase) ->
        Path = [stock, Wood],
        {ok, Stock} = khepri_tx:get(Path),
        NewStock = erlang:min(?MAX_STOCK, Stock + Increase),
        ok = khepri_tx:put(Path, NewStock).

    increase_wood_stock(Wood, Increase) ->
        khepri:transaction(
          ?STORE_ID, fun generic_tx/2, [Wood, Increase]).
    ```

    This way, one can write a more generic transaction function which doesn't rely on variables in its environment, and pass arguments to `khepri:transaction()` instead.

4.  **Stored procedures can be used as transaction functions**. This allows to extract and store a transaction function once and use it many times. This saves the extraction time and reduce the size of the transaction Ra command.

    Here is an example:

    ```erlang
    generic_tx(Wood, Increase) ->
        Path = [stock, Wood],
        {ok, Stock} = khepri_tx:get(Path),
        NewStock = erlang:min(?MAX_STOCK, Stock + Increase),
        ok = khepri_tx:put(Path, NewStock).

    store_wood_stock_tx() ->
        khepri:put(?STORE_ID, [tx, wood_stock], fun generic_tx/2).

    increase_wood_stock(Wood, Increase) ->
        khepri:transaction(
          ?STORE_ID, [tx, wood_stock], [Wood, Increase]).
    ```

    The constraints for transaction functions remain the same: they can't have side effects. So now, everytime a function is stored in Khepri, we verify during extraction if it can be used as a transaction function and store the result with the extracted function. When the stored procedure is executed as a transaction function, we check the result of that verification and possibly deny the execution.

Fixes #141.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -281,7 +281,8 @@
                                               child_list_length |
                                               child_names |
                                               payload |
-                                              has_payload],
+                                              has_payload |
+                                              raw_payload],
                           include_root_props => boolean()}.
 %% Options used during tree traversal.
 %%
@@ -292,7 +293,8 @@
 %% returned tree node properties map. The default is `[payload,
 %% payload_version]'. Note that `payload' and `has_payload' are a bit special:
 %% the actually returned properties will be `data'/`sproc' and
-%% `has_data'/`is_sproc' respectively.</li>
+%% `has_data'/`is_sproc' respectively. `raw_payload' is for internal use
+%% only.</li>
 %% <li>`include_root_props' indicates if root properties and payload should be
 %% returned as well.</li>
 %% </ul>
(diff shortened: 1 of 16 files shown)


src/khepri.erl
```diff
@@ -2275,74 +2289,207 @@ transaction(Fun) ->
...
       Ret :: khepri_machine:tx_ret() | khepri_machine:async_ret().
 %% @doc Runs a transaction and returns its result.
 %%
-%% `Fun' is an arbitrary anonymous function which takes no arguments.
+%% `Fun' is an arbitrary anonymous function which takes the content of `Args'
+%% as its arguments. In other words, the length of `Args' must correspond to
+%% the arity of `Fun'.
+%%
+%% Instead of `Fun', `PathPattern` can be passed. It must point to an existing
+%% stored procedure. The length ot `Args' must correspond to the arity of that
```
the-mikedavis, line 2488:
```suggestion
%% stored procedure. The length of `Args' must correspond to the arity of that
```

src/khepri_payload.erl
```diff
@@ -83,9 +84,15 @@ data(Term) ->
...
 sproc(Fun) when is_function(Fun) ->
-    #p_sproc{sproc = Fun};
+    #p_sproc{sproc = Fun,
+             %% This will be overriden with the correct value when the
+             %% function will be extracted.
+             is_valid_as_tx_fun = false};
 sproc(#standalone_fun{} = Fun) ->
-    #p_sproc{sproc = Fun}.
+    #p_sproc{sproc = Fun,
+             %% This will be overriden with the correct value when the
```
the-mikedavis, line 93: there's a typo on these lines: overriden -> overridden

the-mikedavis (commented):
Based on how I've seen transactions being used in the server, I think this will be a big usability improvement 🎉

src/khepri.erl
dumbbell: Fixed.

src/khepri_payload.erl
dumbbell: Fixed.

dumbbell:
> Based on how I've seen transactions being used in the server, I think this will be a big usability improvement tada

Thank you!

I addressed your comments and force-pushed the changed.

the-mikedavis (approved):

PR #156, opened 2022-10-21
Support empty binaries in string-based paths

dumbbell:
This adds the ability to convert a path such as `[<<>>]` to a string (`"//"` in this case).

URIs can have meaningful consecutive forward slashes. We use this to encode empty binaries in string-based paths. This is breaking change because consecutive forward slashes were ignored before.

Here are a few examples:
* `[<<>>]` = `"//"`
* `[<<>>, <<>>]` = `"///"`
* `[foo, <<>>, bar]` = `"/:foo//:bar"`
* `[foo, bar, <<>>]` = `"/:foo/:bar//`

### DIFF
--- a/src/khepri_path.erl
+++ b/src/khepri_path.erl
@@ -235,7 +235,11 @@ compile(PathPattern) ->
 %%
 %% For convenience, a native path is also accepted and returned as-is.
 
-from_string("/" ++ MaybeString) ->
+from_string([$/, $/ | MaybeString]) ->
+    %% The path starts with two forward slashes. Therefore the path starts
+    %% with an empty binary.
+    from_string([$/ | MaybeString], [<<>>, ?KHEPRI_ROOT_NODE]);
+from_string([$/ | MaybeString]) ->
     from_string(MaybeString, [?KHEPRI_ROOT_NODE]);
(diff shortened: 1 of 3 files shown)


dumbbell:
FTR, we will use empty binaries in RabbitMQ.

src/khepri_path.erl
```diff
@@ -310,6 +314,18 @@ from_string([Char] = Rest, [] = ReversedPath)
...
+    %% Two consecutive forward slashes mean there is an empty binary
+    %% component.
+    ReversedPath1 = prepend_component(<<>>, ReversedPath),
+    from_string([$/ | Rest], ReversedPath1);
+
+%from_string([$/], ReversedPath) ->
+%    %% A path ending with a forward slash means that the path ends with an
+%    %% empty binary component.
+%    ReversedPath1 = prepend_component(<<>>, ReversedPath),
+%    finalize_path([], ReversedPath1);
```
the-mikedavis, line 327: This ends up being covered by the pattern above, right?

the-mikedavis (approved):

src/khepri_path.erl
dumbbell: Indeed, this was left over during development. Removed, thank you!

PR #157, opened 2022-10-21
Add import/export feature

### LINKED ISSUE #153: Provide an import/export feature
An import/export feature would be nice to:
* quickly initialize a new Khepri store
* snapshot or backup a production Khepri store

When I mentionned this as a future Khepri feature on the Erlang forum, [LeonardB suggested me to re-use Mnesia's backup & restore feature](https://erlangforums.com/t/khepri-a-tree-like-replicated-on-disk-database-library-for-erlang-and-elixir-introduction-feedbacks/438/56), so that callback modules could used with both Khepri and Mnesia. I think this is a good idea.

### Implementation ideas

From the [Mnesia backup & restore documentation](https://www.erlang.org/doc/apps/mnesia/mnesia_app_a), a callback module provides the following interface:

```erlang
-export([
         %% Write access
         open_write/1,
         write/2,
         commit_write/1,
         abort_write/1,

         %% Read access
         open_read/1,
         read/1,
         close_read/1
        ]).
```

In other words, both the import (restore) and export (backup) code paths have setup and teardown calls. Between those two calls, `read/1` or `write/2` are called in a loop to import or export chunks of records. A chunk of records is a list of opaque terms: the callback module does not know how a Mnesia (or Khepri) entry is represented. An empty list represents the "end of records".

To export a Khepri store, here are a few notes:
* We could walk the tree (like we do for any operations on the tree in `khepri_machine`) and call `write/2` for every tree node.
* We would have to pay attention to `keep_while` conditions and export them as well.
* We could skip tree nodes which don't carry anything (no payload, no child nodes, no `keep_while` conditions); those would be recreated automatically during import.
* The format of an exported tree node could be a `khepri_machine` `#put{}` command.

Then to import, we could offer the possibility to reset the existing tree nodes before or keep them.

Because we would use the `walk_down_the_tree()` function, we can benefit from conditions and export very specific tree nodes instead of everything:

```erlang
%% Export everything.
khepri:export(StoreId).

%% Export only the wood stock for woods with a name starting with `c'.
khepri:export(StoreId, [stock, wood, #if_name_matches{regex = "^c"}]).
```

### In the future

* If we re-use the Mnesia backup & restore callback module interface, perhaps we can use this to transfer records between Mnesia and Khepri. The callback module would probably need another callback to take care of the conversion of records/payloads between Mnesia and Khepri.
* I wonder if this mechanism could be used to implement a more efficient Ra snapshotting code tailored for Khepri.

dumbbell:
With this patch, it is possible to export and import a Khepri store.

`khepri:export/{2,3,4}` uses a callback module and private data to export the content of the store. It is possible to specify a path pattern to filter what should be exported.

`khepri:import/{2,3}` uses a callback module and private data — of course the same used for export — to import what was previously exported. This function doesn't remove existing tree nodes. It is up to the caller to take care of this. There is no support for filtering on import.

The callback module API is the same as [Mnesia's "Backup and restore" API](https://www.erlang.org/doc/apps/mnesia/mnesia_app_a). The callback module itself doesn't know what is inside the list of "backup items" passed to it, it's completely opaque.

The implementation uses `khepri_machine` state machine commands as the export format: each tree node matching the export path pattern is converted to a `put` command. Tree nodes which don't have any payload or "keep while" conditions are skipped (i.e. not exported).

During import, the commands are processed one by one. In a future patch, we could process them in batches to speed things up.

Fixes #153.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -138,6 +138,9 @@
          'delete_many_payloads!'/1, 'delete_many_payloads!'/2,
          'delete_many_payloads!'/3,
 
+         export/2, export/3, export/4,
+         import/2, import/3,
+
          info/0,
          info/1, info/2]).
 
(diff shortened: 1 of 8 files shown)


dumbbell:
This work in unfinished at this stage:
* It lacks type spec and Dialyzer is probably unhappy for now.
* It lacks documentation.
* Perhaps we want to support batch processing in this pull request as well.

src/khepri.erl
```diff
@@ -2594,8 +2597,162 @@ wait_for_async_ret(Correlation, Timeout) ->
...
+%%        "export-wodd-stock-1.erl").
+%% '''
+%%
+%% @param StoreId the name of the Khepri store.
+%% @param PathPattern the path pattern matching the tree nodes to export.
+%% @param Module the callback module to use to export.
+%% @param ModulePriv arguments passed to `Module:open_write/1'.
+%%
+%% @returns `ok' or an `{ok, Term}' tuple if the export succeeded (the actual
+%% return value depends on wheter the callback module wants to return anything
```
the-mikedavis, line 2693:
```suggestion
%% return value depends on whether the callback module wants to return anything
```

looks like there's one more of these down on L2427

the-mikedavis (commented):
I think this API looks good 👍

I was thinking that the `khepri_export_erlang` importer/exporter might need to use something like `term_to_binary/1`/`binary_to_term/1` for encoding/decoding because of functions in stored procedure but pretty-printing ends up working well because of the functions are just records with the beam code. Could we add a stored procedure case to `test/export.erl` so this is tested too?

the-mikedavis:
> Perhaps we want to support batch processing in this pull request as well

I'm not sure what batch processing will look like. This PR looks well-contained though with just the import/export changes so I think this should be split to a separate PR unless its implementation and import/export are very similar or inform each other.

src/khepri.erl
```diff
@@ -2594,8 +2597,162 @@ wait_for_async_ret(Correlation, Timeout) ->
...
+%% ok = khepri:export(StoreId, khepri_export_erlang, "export-1.erl").
+%% '''
+%%
+%% Example: export a subset of the Khepri store
+%% ```
+%% ok = khepri:export(
+%%        StoreId,
+%%        "/:stock/:wood/**",
+%%        khepri_export_erlang,
+%%        "export-wodd-stock-1.erl").
```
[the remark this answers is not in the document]
dumbbell: Fixed.

src/khepri.erl
dumbbell: Fixed.

dumbbell:
> I think this API looks good +1
> 
> I was thinking that the `khepri_export_erlang` importer/exporter might need to use something like `term_to_binary/1`/`binary_to_term/1` for encoding/decoding because of functions in stored procedure but pretty-printing ends up working well because of the functions are just records with the beam code. Could we add a stored procedure case to `test/export.erl` so this is tested too?

Thank you! Yes, I will add testcases with extracted functions. Some of them could be actual function references. We should probably extract it at the time of export.

dumbbell:
> > Perhaps we want to support batch processing in this pull request as well
> 
> I'm not sure what batch processing will look like. This PR looks well-contained though with just the import/export changes so I think this should be split to a separate PR unless its implementation and import/export are very similar or inform each other.

You are right, let's prepare another patch for batching. Import/export can take advantage of it but doesn't require it.

src/khepri.erl
```diff
@@ -2594,8 +2597,162 @@ wait_for_async_ret(Correlation, Timeout) ->
...
+%% ```
+%% ok = khepri:import(StoreId, khepri_export_erlang, "export-1.erl").
+%% '''
+%%
+%% @param StoreId the name of the Khepri store.
+%% @param Module the callback module to use to import.
+%% @param ModulePriv arguments passed to `Module:open_read/1'.
+%%
+%% @returns `ok' or an `{ok, Term}' tuple if the import succeeded (the actual
+%% return value depends on wheter the callback module wants to return anything
```
the-mikedavis, line 2747:
```suggestion
%% return value depends on whether the callback module wants to return anything
```

dumbbell: Thanks, I missed the fact you mentionned the mistake was in two places.

the-mikedavis (approved):

PR #159, opened 2022-10-31
Implement ra_machine:overview/1 for khepri_machine

the-mikedavis:
The overview callback formats the state of the machine for functions like `sys:get_status/1`. The actual state of the machine is still accessible with `sys:get_state/1`.

The status output was not very readable because of the beam code binaries in `standalone_fun` records and the use of records for nodes in the internal representation of the tree. This `overview/1` implementation uses the map representation of the tree from `khepri_utils:flat_struct_to_tree/1` and converts compiled standalone functions into regular functions with `erlang:make_fun/3` (which are printed with a more concise and readable format).

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -64,6 +64,7 @@
 
 -export([to_standalone_fun/1,
          to_standalone_fun/2,
+         to_fun/1,
          exec/2]).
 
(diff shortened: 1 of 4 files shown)


dumbbell (approved):
Thanks!

PR #160, opened 2022-10-31
Generate docs with ExDoc

the-mikedavis:
ExDoc documentation has some nice out-of-the-box features:

* nice styling
* dark and light modes based on your OS settings
* built-in search
* popups when hovering over items
* a selector for jumping between versions published to HexDocs
* links from docs to the appropriate source code

We can use ExDoc for an Erlang library without requiring an Elixir installation with the [`rebar3_ex_doc`](https://github.com/starbelly/rebar3_ex_doc) plugin. An existing example that follows this is the [`telemetry`](https://github.com/beam-telemetry/telemetry) library: https://hexdocs.pm/telemetry/readme.html

This change integrates that plugin and migrates the existing documentation to ExDoc. This includes two changes:

* `overview.edoc` becomes `overview.md` - ExDoc "extras" are written in Markdown
* Some syntax change for references, especially in `overview.md`:
    * In ExDoc when writing Erlang syntax, types are referred to as `t:module:type_name/arity`.
    * Callbacks are `c:module:callback_name/arity`

Here's a preview: https://the-mikedavis.github.io/khepri/overview.html

---

This is WIP while we wait on some PRs to be released upstream in ExDoc and rebar3_ex_doc.


the-mikedavis:
I'll revisit this once Erlang/OTP 27 becomes available (or when it becomes the minimum version). Erlang/OTP 27 lands [EEP59](https://www.erlang.org/eeps/eep-0059) which proposes moving towards markdown and ExDoc and away from EDoc (specifically see https://www.erlang.org/eeps/eep-0059#what-about-edoc-and-erl_docgen)

PR #161, opened 2022-11-02
Fix `call_fun2` instruction jump label support

dumbbell:
Since #149, we resolve and adjust the jump label inside the `call_fun2` instruction. This way we are as close as possible to the code originally produced by the compiler and benefit from its optimizations.

However, there is one situation where we can't keep the jump label: it is when the called local function is not local anymore after extraction.

To avoid many almost identical copies of the same anonymous function where only the environment changes, `khepri_fun` will extract the function in a dedicated module and keep the environment separate. If there is another anonymous function inside this environment, it will become its own module.

Therefore, it's possible that in the original Beam code, the two anonymous functions were inside the same module. If the compiler emitted a `call_fun2` instruction, it could use a jump label in the outer function to call the inner function passed in the environment.

After extraction by `khepri_fun`, this is no longer possible because the inner function will be in another module. `call_fun2` must use the `{atom, safe}` argument instead.

While here, add support for the `{atom, unsafe}` argument as well because I found a testcase for it while searching for the testcase for the jump label version.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -2298,8 +2298,20 @@ pass2_process_instruction(
        BsPutSomething =:= bs_put_integer ->
     replace_label(Instruction, 2, State);
 pass2_process_instruction(
-  {call_fun2, {f, _}, _, _} = Instruction, State) ->
-    replace_label(Instruction, 2, State);
+  {call_fun2, {f, _} = Label, _, _} = Instruction,
+  #state{label_map = LabelMap} = State) ->
+    %% If the label used in this instruction is unavailable in the label map,
+    %% it means the targeted function was not extracted with this one. It
+    %% must be that the called function is another anonymous function inside
+    %% the same module. However, after extraction, both this function and the
+    %% called one end up in two separate modules.
+    %%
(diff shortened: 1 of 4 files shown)


src/khepri_fun.erl
```diff
@@ -2298,8 +2299,20 @@ pass2_process_instruction(
        BsPutSomething =:= bs_put_integer ->
     replace_label(Instruction, 2, State);
 pass2_process_instruction(
-  {call_fun2, {f, _}, _, _} = Instruction, State) ->
-    replace_label(Instruction, 2, State);
+  {call_fun2, {f, _} = Label, _, _} = Instruction,
+  #state{label_map = LabelMap} = State) ->
+    %% If the label used in this instruction is unavailable in the label map,
+    %% it means the targetted function was not extracted with this one. It
```
the-mikedavis, line 2305:
```suggestion
    %% it means the targeted function was not extracted with this one. It
```

the-mikedavis (approved):
Nice catch! 🎣

src/khepri_fun.erl
dumbbell: Thanks, fixed

PR #163, opened 2022-11-07
Add `maps:fold/3`-like functions

dumbbell:
This patch adds the following new public functions:
* `khepri:fold/5`
* `khepri:foreach/4`
* `khepri:map/4`
* `khepri:filter/4`

The internals of the state machine implementation is already based on a walk of the tree, using a callback function to handle generic queries and updates. `khepri_machine:get/3` is replaced by `khepri_machine:fold/5` to expose this to the rest of the library. This allows to implement `khepri:fold/5`. Other folding-like functions are based on `khepri:fold/5`.

The same functions are also available in `khepri_tx`.


src/khepri.erl
```diff
@@ -1278,7 +1317,448 @@ count(PathPattern, Options) when is_map(Options) ->
...
+%% The returned `{ok, NewAcc}' tuple contains the return value of the last call
+%% to `Fun', or `Acc' if no tree nodes matched the given path pattern.
+%%
+%% Example: list all nodes in the tree
+%% ```
+%% %% List all tree node paths in the tree. The tree is:
+%% %% <root>
+%% %% `-- foo
+%% %%     `-- bar = value
+%% {ok, [[foo], [foo, bar]]}} = khepri:fold(
```
the-mikedavis, line 1416:
```suggestion
%% {ok, [[foo], [foo, bar]]} = khepri:fold(
```

src/khepri.erl
```diff
@@ -1278,7 +1317,448 @@ count(PathPattern, Options) when is_map(Options) ->
...
+%%                                  StoreId,
+%%                                  fun(Path, _NodeProps, Acc) ->
+%%                                      [Path | Acc]
+%%                                  end, [],
+%%                                  [?KHEPRI_WILDCARD_STAR_STAR]).
```
the-mikedavis, line 1421: I think the PathPattern and Fun arguments should be swapped here, right?

the-mikedavis: It looks like the same applies to the code examples for the other new functions too

src/khepri.erl
```diff
@@ -1278,7 +1317,448 @@ count(PathPattern, Options) when is_map(Options) ->
...
+%% </ol>
+%%
+%% Example: only keep tree nodes under `/:foo'
+%% ```
+%% %% The tree is:
+%% %% <root>
+%% %% `-- foo
+%% %%     `-- bar = value
+%% {ok, #{[foo] => undefined,
+%%        [foo, bar] => value} = khepri:filter(
```
the-mikedavis, line 1730:
```suggestion
%%        [foo, bar] => value}} = khepri:filter(
```

src/khepri.erl
dumbbell: Fixed, thanks!

src/khepri.erl
dumbbell: Fixed, thanks!

src/khepri.erl
dumbbell: You're right, thanks!

I miss the testability of code snippets in the documentation, like with Rust.

the-mikedavis (approved):

dumbbell:
Thanks @the-mikedavis for the approval! I will wait for #135 to be merged, otherwise it will create yet another conflict for you. This patch will be easier to adapt instead.

PR #165, opened 2022-11-07
khepri_fun: Add guard for unwrapping type-tagged registers

the-mikedavis:
The latest `beam_disasm` (OTP26+) unwraps the type-tag. Registers passed into this function after that change are mistakenly unwrapped, so types like

```erl
#tr{reg = {x, 0}, t = #t_map{super_key = any, super_value = any}}
```

Are mistakenly unwrapped as

```erl
#tr{reg = {x, 0}, t = t_map}
```

Which causes an internal failure in `beam_validator` when passed into `compile:forms/2`.

This change prevents this future breakage by guarding on the type being a tuple. This match should never succeed on OTP26+ since all types in `beam_types.hrl` are specified as records, so the first element will always be an atom.


dumbbell (approved):
Thanks!

Issue #144, opened 2022-10-03
Multiple inserts in a row are slow because of the snapshotting


dumbbell:
When we import definitions in RabbitMQ, we perform many `#put{}` in a row. Currently, Khepri's state machine asks Ra to take a snapshot every 4096 commands. With a large definitions import, this causes a lot of snapshotting.

The first suggestion from @kjnilsson was to skip snapshotting when inserting new tree nodes. Indeed, attempting a snapshot after this won't free anything: new tree nodes inserted don't replace other tree nodes. Therefore, if we don't count new tree nodes in the threshold towards the next snapshot, we might save time.

dumbbell:
This should be fixed after #392.

PR #167, opened 2022-11-08
khepri_cluster: Accept `machine_config` in Ra server config

dumbbell:
It is now possible to pass state machine arguments in the Ra server config. This is handy to e.g. set the snapshot interval:

```erlang
RaServerConfig = #{cluster_name => StoreId,
                   machine_config => #{snapshot_interval => 10000}},
khepri:start(RaSystem, RaServerConfig).
```


dumbbell:
This helps with #144 but doesn't solve it magically.

the-mikedavis (approved):

PR #168, opened 2022-11-08
khepri_fun: Get register from type-tag for call_fun2 translation

the-mikedavis:
OTP26 unwraps the min and max information out of the `tr` record's type field, causing this pattern match to fail. Instead we can use the `get_reg_from_type_tagged_beam_register/1` helper to retrieve the register regardless of the type.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -1170,10 +1170,7 @@ pass1_process_instructions(
     Instruction = {call_fun2, {atom, SafeOrNot}, Arity, {tr, FunReg, Type}},
     pass1_process_instructions([Instruction | Rest], State, Result);
 pass1_process_instructions(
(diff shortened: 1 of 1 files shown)


the-mikedavis:
This removes the part of the pattern match that asserts that the type is a `t_fun`. I think it's safe to ignore the type, especially since this branch replaces it anyways. If the compiler produced a `call_fun2` then the `Func` register must point to a `t_fun` type, right?

Alternatively we could add as assertion on the type or add a new clause that matches on a properly formed `tr` record. WDYT?

dumbbell (approved):
I agree with your analysis. I don't know if we can get something else than `t_fun`. But as you said, we replace it anyway. I hope the compile will reject the generated instruction if we replaced something else by a `t_fun` :-)

PR #170, opened 2022-11-10
khepri_fun: Fix translation for bs_get_float2

the-mikedavis:
The resolution for `bs_get_float2` in `beam_disasm` is [in the same block and has the same body](https://github.com/erlang/otp/blob/7da89b77373985277892ab8b957550be5ca2474d/lib/compiler/src/beam_disasm.erl#L1042-L1050) as the translations for `bs_get_integer2` and `bs_get_binary2` so it has the same mistranslations and can be fixed in the same way.

We can provoke a `{test,bs_get_float2,..}` instruction by pattern matching on a bitstring that uses the `float` type specifier.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -1205,7 +1205,8 @@ pass1_process_instructions(
   State,
   Result)
   when (BsGetSomething =:= bs_get_integer2 orelse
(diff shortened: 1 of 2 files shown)


dumbbell (approved):
Thanks!

PR #176, opened 2022-12-09
khepri_fun: Work around Rebar's use of `cth_readable` parse_transform

dumbbell:
To improve the output of common_test in the terminal, Rebar uses [`cth_readable`](https://hex.pm/packages/cth_readable). This library comes with a parse_transform module to convert all calls to `ct:pal()`.

The application modules are thus recompiled on-the-fly (i.e. the result is not written to disk) with this parse_transform. The result is that the module loaded in memory doesn't match the one on disk (different checksums).

This causes the following exception while trying to extract a transaction function. Here is an example:

```erlang
{khepri_ex, failed_to_prepare_tx_fun,
 #{error =>
   {mismatching_module_checksum,
    #{checksum_from_fun =>
      <<194,208,71,94,61,101,247,31,220,109,126,242,101,120,40,129>>,
      checksum_on_disk =>
      <<152,39,250,105,94,89,145,46,251,4,237,114,45,132,63,33>>,
      compile_info_from_fun =>
      [{version, "8.2.1"},
       {options,
        [{d, 'COMMON_TEST'},
         {d, 'TEST'},
         {i, ".../_build/test/lib/khepri_mnesia_migration/src"},
         {i, ".../_build/test/lib/khepri_mnesia_migration/test"},
         {i, ".../_build/test/lib/khepri_mnesia_migration/include"},
         {i, ".../_build/test/lib/khepri_mnesia_migration"}]},
       {source, ".../src/m2k_table_copy.erl"}],
      compile_info_on_disk =>
      [{version, "8.2.1"},
       {options,
        [debug_info,no_spawn_compiler_process,
         {d, 'COMMON_TEST'},
         warn_export_vars,
         {d, 'TEST'},
         {parse_transform,cth_readable_transform},
         {i, ".../_build/test/lib/khepri_mnesia_migration/src"},
         {i, ".../_build/test/lib/khepri_mnesia_migration/test"},
         {i, ".../_build/test/lib/khepri_mnesia_migration/include"},
         {i, ".../_build/test/lib/khepri_mnesia_migration"}]},
       {source, ".../src/m2k_table_copy.erl"}],
      module => m2k_table_copy}}}}
```

The solutions consists in comparing the source filename and compiler options. If both copies are compiled for tests and one of them uses `cth_readable_transform`, we accept the inconsistency and continue.

While here, the error returned when modules mismatch was updated to include more details.

### DIFF
--- a/src/khepri_fun.erl
+++ b/src/khepri_fun.erl
@@ -72,7 +72,8 @@
          override_object_code/2,
          get_object_code/1,
          decode_line_chunk/2,
-         compile/1]).
+         compile/1,
+         to_actual_arg/1]).
 -endif.
 
 %% FIXME: compile:forms/2 is incorrectly specified and doesn't accept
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):

dumbbell:
Thank you @the-mikedavis for the quick review!

I can't reproduce the problem outside of `khepri_mnesia_migration`. I would prefer to come up with a testcase for this patch. That's why I'm not ready to commit it yet.

PR #180, opened 2023-01-23
Bundle keep-while metadata and tree root into a tree record

the-mikedavis:
This is an internal refactor for `khepri_machine` that bundles together the keep-while metadata (`keep_while_conds` and `keep_while_conds_rev_idx` on the `khepri_machine` record) together with the tree `root`. These three values are often read and written to together by functions that walk across or modify the tree.

The first commit creates the tree record and the next replaces separate uses of `Root` and the keep-while metadata with the tree record. This eliminates the `Extra` argument and type which merged keep-while metadata with `AppliedChanges`.


src/khepri_machine.erl
```diff
@@ -284,7 +282,7 @@ delete(StoreId, PathPattern, Options) when ?IS_STORE_ID(StoreId) ->
     PathPattern1 = khepri_path:from_string(PathPattern),
     khepri_path:ensure_is_valid(PathPattern1),
     {CommandOptions, TreeOptions} = split_command_options(Options),
-    %% TODO: Ensure `Extra' is unset.
+    %% TODO: Ensure `PutOptions' are not set included this map.
```
dumbbell, line 285: "(...) are not set in this map."?

the-mikedavis: Whoops good catch 😅

dumbbell (approved):

PR #183, opened 2023-02-15
khepri_tx: Support new Erlang/OTP 26 instructions

the-mikedavis:
This change fixes compatibility with Erlang/OTP 26.0-rc1 release candidate which was cut today: https://github.com/erlang/otp/releases/tag/OTP-26.0-rc1

There are two new instructions: `bs_match/3` and `update_record/5`. `bs_match/3` optimizes some bitstring matching and `update_record/5` is used to update record tuples.

`bs_match/3` replaces some other bitstring matching instructions which are already present in the test suite, so this instruction is already covered while running on Erlang/OTP 26. `update_record/5` can be provoked with a new test case which updates a record field.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,11 +11,14 @@ jobs:
 
     strategy:
       matrix:
-        otp_version: [24, 25]
+        otp_version: [24, 25, 26.0-rc1]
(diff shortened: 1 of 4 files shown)


michaelklishin:
@the-mikedavis man you are fast :)

dumbbell (approved):

PR #192, opened 2023-03-28
Extract `khepri_fun` and create the Horus library

dumbbell:
### Why
`khepri_fun` is already a complex piece of code. Its mission is not exactly specific to Khepri and it would be easier to maintain and test as a standalone project.

### How
A new library was created out of `khepri_fun` called Horus. See: https://github.com/rabbitmq/horus


test/cluster_SUITE.erl
the-mikedavis, line 1: Could the changes here be split into a separate change/PR? I think I see the fix for this intermittent failure also: we should be calling `rpc:call` on `Node` rather than `Node1` on L1255 (old) / L1259 (new)

the-mikedavis: I pushed 295bf1e91a20e5260ee3df9fdb8423b16c1d518a which I think should fix it (🤞) so I think the changes in this file can be dropped now. I'll re-run that locally in a loop for a while to see if it can still break.

the-mikedavis (approved):

PR #208, opened 2023-05-26
Remove debug logs for successful query of members

dcorbacho:
It just floods the logs on any long run. Just leave the failure ones.


dumbbell:
Because it is not a cheap call, I wanted to keep them around until we have something lighter to query cluster members. What do you think? I say that, but I don't have an implementation idea yet.

dcorbacho:
If you have any use for the logs, just leave them. But the issue I have is not with the query itself, but the huge logs it generates where this repeated line just becomes noise. It needs to be filtered out every time to find the relevant information on the logs.

dumbbell (approved):

PR #217, opened 2023-07-26
khepri_cluster: Make `join/1` idempotent

dumbbell:
### Why

If `khepri_cluster:join/1` is called twice in a row, the join succeeds because the joining node is reset before it joins the cluster again. This is not exactly idempotent, but the end result is the same.

However if the joining node was reset somehow (e.g. its data directory was lost) and therefore thinks it is a standalone store, calling `khepri_cluster:join/1` again would return `{error, already_member}`.

It is more comfortable if the function just does what it takes to achieve the wanted result of expanding the cluster.

### How

The `{error, already_member}` error is caught and the function returns `ok` instead. Behind the scene, Ra did the correct thing and the joining node is back in the cluster.

### DIFF
--- a/src/khepri_cluster.erl
+++ b/src/khepri_cluster.erl
@@ -761,6 +761,12 @@ do_join_locked(StoreId, ThisMember, RemoteNode, Timeout) ->
                "Cluster for store \"~s\" successfully expanded",
                [StoreId]),
             ok;
+        {error, already_member} ->
+            ?LOG_DEBUG(
+               "This node (~0p) is already a member of the remote node's "
+               "cluster (~0p)",
+               [ThisMember, RemoteMember]),
+            ok;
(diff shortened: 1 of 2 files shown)


src/khepri_cluster.erl
the-mikedavis, line 1: What do you think about adding a small note to the `join/3` docs?

dumbbell: Good idea!

dumbbell: Done :-)

the-mikedavis (approved):

dumbbell:
GitHub merged the initial commit (7ba03ec1ca27293438ddc996d4da746dcfdbff64), not the updated one (ff9060ab98e9e8614c4e0fd9a2b7ab8a7331248f)... We need to revert that pull request and create a new one :-/

Issue #213, opened 2023-07-20
Lack of mechanism to update or delete projection once it has been created


xh4:
### Why

I noticed that once a projection has been registered, register it again with a new projection function has no effects.

The following code shows that `register_projection` won't update the projection tree if the ETS table already exists.

```erlang
%% khepri_machine.erl (L1143-L1156)
apply(
  Meta,
  #register_projection{pattern = PathPattern, projection = Projection},
  #?MODULE{tree = Tree, projections = ProjectionTree} = State) ->
    Reply = khepri_projection:init(Projection),
    State1 = case Reply of
                 ok ->
                     restore_projection(Projection, Tree, PathPattern),
                     ProjectionTree1 = khepri_pattern_tree:update(
                                         ProjectionTree,
                                         PathPattern,
                                         fun (?NO_PAYLOAD) ->
                                                 [Projection];
                                             (Projections) ->
                                                 [Projection | Projections]
                                         end),
                     erase(compiled_projection_tree),
                     State#?MODULE{projections = ProjectionTree1};
                 _  ->
                     State
             end,
    Ret = {State1, Reply},
    bump_applied_command_count(Ret, Meta).
```

Also, it seems that there is no easy way to delete a projection once it has been created.

### How

- Add code to update projection tree even though ETS table exists.
- Provide functions to unregister projection and delete corresponding ETS table.

PR #220, opened 2023-08-02
Add a command to unregister projections

the-mikedavis:
Projections can be registered but not removed once they're in the store. This change adds functions to remove projections from the store by name, so if we need to alter projections in the future we can call 'unregister_projection' and then register replacements.

See https://github.com/rabbitmq/khepri/issues/213


src/khepri.erl
```diff
@@ -2929,6 +2931,75 @@ register_projection(StoreId, PathPattern, Projection, Options) ->
...
+      Options :: khepri:command_options(),
+      Ret :: ok | khepri:error().
+%% @doc Removes a projection from the store by name.
+%%
+%% This function accepts the following two forms:
+%% <ul>
+%% <li>`unregister_projection(StoreId, ProjectionName)'. Calling it is the same
+%% calling `unregister_projection(StoreId, ProjectionName, #{})'.</li>
+%% <li>`unregister_projection(ProjectionName, Options)'. Calling it is the same
+%% calling `unregister_projection(StoreId, ProjectionName, Options)' with the
```
dumbbell, line 2969: Shouldn't it say "the same **as** calling"?

dumbbell (approved):

PR #223, opened 2023-08-16
khepri_machine: Handle `{error,shutdown}` response from Ra queries

the-mikedavis:
Ra's query functions may return `{error,shutdown}` if the server process receives the query but shuts down before it is able to reply. This is the same as the change in #211 but covers queries.


dumbbell (approved):
Thank you!

PR #226, opened 2023-09-27
khepri_machine: Run the query anonymous function from the calling process

dumbbell:
### Why

Before this change, the anonymous function passed to `khepri_machine:fold/5` was executed by the Ra server process. This had two downsides:
* it blocked the Ra server to serve writes
* the anonymous function couldn't query the Ra server itself because it would deadlock

### How

The alternative implemented in this patch is to get the whole Khepri tree from the Ra server, then run the anonymous function, this time from the calling process.

The downside is that the whole Khepri tree is copied to that calling process.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -303,13 +303,19 @@
 %% </ul>
 
 -type query_options() :: #{timeout => timeout(),
-                           favor => favor_option()}.
+                           favor => favor_option(),
+                           copy_tree_and_run_from_caller => boolean()}.
 %% Options used in queries.
 %%
 %% <ul>
 %% <li>`timeout' is passed to Ra query processing function.</li>
 %% <li>`favor' indicates where to put the cursor between freshness of the
 %% returned data and low latency of queries; see {@link favor_option()}.</li>
+%% <li>`copy_tree_and_run_from_caller' indicates if the query anonymous
+%% function should run from the calling process instead of the Ra process.
+%% This allows to recursively query the Khepri store from the anonymous
+%% function. However, this means the Khepri tree is copied to the calling
+%% process. Defaults to `false'</li>
 %% </ul>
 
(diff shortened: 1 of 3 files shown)


dumbbell:
I'm still not sure this alternative approach is what we want. Copying the entire Khepri tree to the calling process could be very expensive and the memory footprint could sky-rocket.

Another solution would be to try to run the anonymous function from the Ra process like before. But if it exits with a `calling_self` exception, then we could re-run it from the calling process. But then, the anonymous function must have no side effects, or the side effects should be idempotent. Perhaps we could accept a `{can_be_restarted, true}` option, so that the caller lets Khepri know it can re-run the anonymous function.

What do you think, @dcorbacho and @the-mikedavis?

To give some context to this issue, when RabbitMQ queries transient queues from the database, it queries the Khepri members from a Khepri query anonymous function:

```erlang
khepri:fold(
  StoreId,
  PathPattern,
  fun(QueuePath, NodeProps) ->
      ...
      %% This queries the Ra server and can't be executed by the Ra server
      %% itself because it would deadlock. gen_statem throws a `calling_self`
      %% exception.
      Members = khepri:members(StoreId),
      ...
  end,
  []).
```

dumbbell:
I pushed an additional commit to implement that option that lets the caller decide if it wants the tree to be copied to be able to do nested queries.

I would love some feedback on the approach and the option name :-)

the-mikedavis:
We ended up going a different direction on this in the server similar to https://github.com/the-mikedavis/rabbitmq-server/commit/aed8ddd80e35d97b1afeb277f057209d845ef548. We use `khepri:get_many/3` to get all of the records under the path and then fold separately with `maps:fold/3` so that the fun isn't called within the Ra server process.

@dumbbell should we close out this PR for now? We could re-open it if we find another use-case and want to make it possible to do while still calling `khepri:fold/5`

PR #229, opened 2023-10-16
Replace `wait_for_async_ret` with `handle_async_ret` for handling async Ra events

the-mikedavis:
This is similar to `wait_for_async_ret/1,2` except that the caller is expected to `receive` the Ra event. This is necessary because the Ra event sent when a batch of commands is applied includes a list of correlation and result pairs, making it impossible to use a selective receive to look for a single correlation ID.

We could alternatively map the `ra_event` message into some other value like `{ok, Correlations} | {error, CorrelationId}` but this isn't much easier to work with than the `ra_event`s themselves and would be brittle for changes to `ra_event`. Instead we leave it to the caller to receive and handle the events. This function is then responsible only for calling internal functions to update the cached leader for the given store ID.

I've also added deprecations for `wait_for_async_ret/1,2` since it can't be used to reliably find if a given correlation ID was applied. Instead we should prefer to have the caller `receive` the Ra event and handle the correlation IDs within.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -125,6 +125,7 @@
          transaction/1, transaction/2, transaction/3, transaction/4,
          transaction/5,
 
+         handle_async_ret/1, handle_async_ret/2,
          wait_for_async_ret/1, wait_for_async_ret/2,
 
          %% Bang functions: they return the value directly or throw an error.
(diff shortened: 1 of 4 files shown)


src/khepri.erl
```diff
@@ -174,6 +175,11 @@
 
                            unwrap_result/1]}).
 
+-deprecated({wait_for_async_ret, 1, "use khepri:handle_async_ret/2 and "
+                                    "`receive' Ra events instead"}).
+-deprecated({wait_for_async_ret, 2, "use khepri:handle_async_ret/2 and "
+                                    "`receive' Ra events instead"}).
+
```
dumbbell, line 182: Given we didn't cut version 1.0.0 yet, we don't really need to follow a deprecation period. I think we can just get rid of these functions.

src/khepri.erl
```diff
@@ -3292,6 +3298,42 @@ transaction(FunOrPath, Args, ReadWrite, Options)
 transaction(StoreId, FunOrPath, Args, ReadWrite, Options) ->
     khepri_machine:transaction(StoreId, FunOrPath, Args, ReadWrite, Options).
 
+-spec handle_async_ret(StoreId, RaEvent) -> ok when
+      StoreId :: khepri:store_id(),
+      RaEvent :: ra_server_proc:ra_event().
```
dumbbell, line 3303: Could you please add a `handle_async_ret/1` variant which uses the default store ID to follow the same pattern as other public functions?

dumbbell (changes requested):

dumbbell (approved):

PR #240, opened 2024-01-05
khepri_cluster: Use `ra_leaderboard` first to query members locally

dumbbell:
## Why

A call to `ra_leaderboard:lookup_members/1` will be faster if it has the answer compared to a Ra server query.

## How

If it doesn't know the list of members, we fall back to the Ra server query.


kjnilsson:
Ra 2.7.2 is released with the ra leaderboard fixes.

https://hex.pm/packages/ra

PR #242, opened 2024-01-12
Always pass exported functions to Ra queries

### LINKED ISSUE #238: Passing `fun()` to `ra:*_query()` may break when executed on a remote node
Passing an anonymous function reference for execution on a remote node may break if the remote node has another version of Erlang or the module hosting the anonymous function.

It would be better to call an exported function.

Reporter by: @ansd

dumbbell:
## Why

Using `fun()` is fragile if it has to be executed on a remote node. Indeed, the remote node may have another version of the module hosting the function and the function reference may be invalid.

## How

There are two parts:

1. We always use an exported function inside Khepri.
2. We ensure that the caller of folding APIs passes MFA if they request leader or consistent queries. `fun()` are only accepted for local queries.

Fixes #238.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -72,6 +72,7 @@
 -include("include/khepri.hrl").
 -include("src/khepri_cluster.hrl").
 -include("src/khepri_error.hrl").
+-include("src/khepri_mfa.hrl").
 -include("src/khepri_ret.hrl").
 
 -export([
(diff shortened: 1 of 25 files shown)


ansd (approved):

dumbbell:
The patch is finished but I would like to test it with RabbitMQ before merging it.

dumbbell:
I have second thoughts on the fact that MFAs are good enough as a solution to execute code remotely. Yes, with the current patch, we ensure that a query won’t break because the function reference is invalid on another node. But that doesn’t mean that the function exported remotely is the same as the local one. This is quite unintuitive.

Thinking out loud here; if a leader query was requested, what about we check the cursor on the leader, wait for the local cursor to catch up, then execute the query function locally? Don’t we have the same consistency guaranty? It may take more time to answer the query though, because the catch up can be longer than a remote execution.

dumbbell:
This pull request was superseded by #260.

PR #250, opened 2024-02-20
khepri_machine: Add command deduplication mechanism

dumbbell:
## Why

The "client" side of `khepri_machine` implemented in `process_sync_command/3` have a retry mechanism if `ra:process_command/3` returns an error such as `noproc`, `nodedown` or `shutdown`.

However, this retry mechanism can't tell if the state machine already received the command and just couldn't reply, for instance because there is a node stopping or a change of leadership.

Therefore, it's possible that the same command is submitted twice and thus processed twice.

That's ok for idempotent commands, but it may not be alright for all transactions for example. That's why we need a deduplication mechanism that ensures the same command is not applied multiple times.

## How

Two new commands are introduced to implement the deduplication system:
* `#dedup{}` which is used to wrap the command to protect and assign a unique reference to it
* `#dedup_ack{}` which is used at the end of the retry loop to let the state machine know that the "client" side received the reply

When the state machine receives a command wrapped into a `#dedup{}` command, it will remember the reply for the initial processing of that command. For any subsequent copies of the same `#dedup{}` (based on the unique reference), the state machine will not apply the wrapped command and will simply returned the reply it remembered from the first application.

Later when the state machine receives a `#dedup_ack{}`, it will drop the cached reply for that reference.

Just in case the client never sends a `#dedup_ack{}`, the state machine will drop any expired cached entries. The expiration time is based on the command timeout. If it's infinity, it defaults to 15 minutes.

This whole deduplication mechanism can be enabled or disabled through the new `protect_against_dups` command option which takes a boolean. This option is off by default, except for R/W transactions.

Thus if the caller knows the transation is idempotent, it can decide to turn the dedup mechanism off.

**V2**: We now use the `effective_machine_version` counter provided by `ra_counters:counters/2` if it is available as it is faster than querying the Ra server. If the counter is unavailable, we fall back to the query. The new counter is added by rabbitmq/ra#426 and will be used once a Ra release contains this change.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -16,6 +16,25 @@
 %% This module is private. The documentation is still visible because it may
 %% help understand some implementation details. However, this module should
 %% never be called directly outside of Khepri.
+%%
+%% == State machine history ==
+%%
+%% <table>
+%% <tr><th>Version</th><th>What changed</th></tr>
+%% <tr>
+%% <td style="text-align: right; vertical-align: top;">0</td>
+%% <td>Initial version</td>
+%% </tr>
+%% <tr>
+%% <td style="text-align: right; vertical-align: top;">1</td>
+%% <td>Added deduplication mechanism:
+%% <ul>
+%% <li>new command option `protect_against_dups'</li>
+%% <li>new commands `#dedup{}' and `#dedup_ack{}'</li>
+%% <li>new state field `dedups'</li>
+%% </ul></td>
+%% </tr>
+%% </table>
 
 -module(khepri_machine).
 -behaviour(ra_machine).
@@ -51,7 +70,8 @@
          apply/3,
          state_enter/2,
          overview/1,
-         version/0]).
+         version/0,
+         which_module/1]).
 
 %% For internal use only.
 -export([clear_cache/1,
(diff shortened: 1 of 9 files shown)


dumbbell:
I need to bump the state machine version and convert the state.

dumbbell:
The pull request is ready for review. The machine version is taken into account, so if the machine is too old, the public API won’t use the dedup mechanism.

src/khepri_machine.erl
```diff
@@ -1400,8 +1554,36 @@ overview(State) ->
...
+
+effective_version(StoreId) ->
+    ThisNode = node(),
+    RaServer = khepri_cluster:node_to_member(StoreId, ThisNode),
+    case ra:member_overview(RaServer) of
+        {ok, #{effective_machine_version := EffectiveMacVer}, _LeaderId} ->
+            {ok, EffectiveMacVer};
+        {error, _} = Error ->
+            Error
+    end.
```
the-mikedavis, line 1586: Once we update Ra to a version that includes https://github.com/rabbitmq/ra/pull/426 I believe this should become:

```suggestion
    case ra_counters:counters(RaServer, [effective_machine_version]) of
        #{effective_machine_version := EffectiveMacVer} ->
            {ok, EffectiveMacVer};
        _ ->
            {error, ?khepri_error(effective_machine_version_not_defined,
                                  #{ra_server => RaServer})}
    end.
```

dumbbell: What about calling `ra:member_overview/1` as a fallback if the counter doesn't provide the version?

the-mikedavis: The counter is set up when the server process initializes so there should be no case where a `ra:member_overview/1` would return successfully while a counters lookup fails

dumbbell: One more thing: I believe we don't need to wait for a new release of Ra that contains the new counter, the code should be compatible with older versions of Ra, shouldn't it?

the-mikedavis: Ah yep. If we fall back to using the `ra:member_overview/1` then we wouldn't have to depend on the most recent changes in Ra :+1:

dumbbell: It works locally when used with an old version of Ra. I committed your suggestion with the fallback to `ra:member_overview/1`.

the-mikedavis (approved):

PR #255, opened 2024-03-19
Use effects to initialize new projections

the-mikedavis:
`khepri_machine` shouldn't initialize projections in its `apply/3` callback. That function might be called by Ra when recovering and folding through the log from a snapshot. If the log contains a `#register_projection{}` command then Khepri will mistakenly create the ETS table and fill it with whatever is in the tree at that moment. Effects like `#trigger_projection{}` are discarded during recovery, so restarting a server could lead to inconsistencies in projection tables between cluster members.

Instead we can use an aux effect to perform the side effects like we do already for `restore_projections` and `#trigger_projection{}` effects. This will be correctly discarded when Khepri is recovering. Then projections will be correctly restored once the machine is recovered with the existing `state_enter(recovered, State)` callback implementation.

To make the `#register_projection{}` `apply/3` clause pure I've added a helper `khepri_machine:has_projection/2`. Exposing it as `khepri:has_projection/2` isn't strictly necessary for this fix but it could be a nice win to take advantage of in the server. Currently we submit `#register_projection{}` commands every time the broker starts and rely on idempotency. A large projection function could be a non-trivial amount of data to store in the log though, so it would be nice to skip the command if we can detect that the projection is already registered.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -76,6 +76,7 @@
          get_triggers/1,
          get_emitted_triggers/1,
          get_projections/1,
+         has_projection/2,
          get_metrics/1]).
 -ifdef(TEST).
 -export([make_virgin_state/1]).
@@ -1906,6 +1907,25 @@ set_emitted_triggers(#khepri_machine{} = State, EmittedTriggers) ->
 
 get_projections(#khepri_machine{projections = Projections}) ->
     Projections.
+
+-spec has_projection(ProjectionTree, ProjectionName) -> boolean() when
+      ProjectionTree :: khepri_machine:projection_tree(),
+      ProjectionName :: atom().
+%% @doc Determines if the given projection tree contains a projection.
+%%
+%% Two projections are considered equal if they have the same name.
+%%
+%% @private
+
+has_projection(ProjectionTree, Name) when is_atom(Name) ->
+    khepri_pattern_tree:any(
+      ProjectionTree,
+      fun(Projections) ->
+              lists:any(
+                fun(#khepri_projection{name = N}) ->
+                        N =:= Name
+                end, Projections)
+      end).
 
 -spec set_projections(State, Projections) -> NewState when
       State :: khepri_machine:state(),
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -1122,6 +1122,13 @@ handle_aux(
                Projection <- Projections]
       end),
     {no_reply, AuxState, LogState};
+handle_aux(
+  _RaState, cast,
+  #restore_projection{projection = Projection, pattern = PathPattern},
+  AuxState, LogState, State) ->
+    Tree = get_tree(State),
+    ok = restore_projection(Projection, Tree, PathPattern),
+    {no_reply, AuxState, LogState};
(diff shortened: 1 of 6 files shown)


src/khepri_machine.erl
```diff
@@ -1207,27 +1215,29 @@ apply(
...
-                     erase(compiled_projection_tree),
-                     set_projections(State, ProjectionTree1);
-                 _  ->
-                     State
-             end,
-    Ret = {State1, Reply},
-    bump_applied_command_count(Ret, Meta);
+    case has_projection(ProjectionTree, ProjectionName) of
+        true ->
+            Ret = {State, {error, exists}},
```
dumbbell, line 1222: Could you please use the `?khepri_error(Name, Info)` macro to create the error and give more context?

Probably something like that, to follow other error patterns:
```erlang
Info = #{name => ProjectionName},
Reason = ?khepri_error(projection_already_exists, Info),
{error, Reason}.
```
I'm just not sure about the name `projection_already_exists`.

dumbbell: That said, do we want to reject an update of a projection?

src/khepri_machine.erl
```diff
@@ -1207,27 +1215,29 @@ apply(
...
+        false ->
+            ProjectionTree1 = khepri_pattern_tree:update(
+                                ProjectionTree,
+                                PathPattern,
+                                fun (?NO_PAYLOAD) ->
+                                        [Projection];
+                                    (Projections) ->
+                                        [Projection | Projections]
+                                end),
+            erase(compiled_projection_tree),
```
dumbbell, line 1233: I didn't pay attention to this line before. Could you please add a comment to explain what is the `compiled_projection_tree` and it is set?

test/projections.erl
dumbbell, line 1: Could you please add a few more testcases to cover all new lines if possible? Look at the coverage report on Codecov. For instance, calls with the default store and calls with options are untested. Likewise for the cases where a projection is added to the list of projections for a given path pattern.

dumbbell (changes requested):

src/khepri_machine.erl
the-mikedavis: That sounds good to me. I believe we already handle the `{error, exists}` in the server so we will need to be careful to change that when we update. I would also like for the server to use the new `has_projection` function though rather than relying on idempotency, since submitting the compiled projection functions each time can only inflate Khepri's Ra log.

`projection_already_exists` sounds good to me :+1:

the-mikedavis: > do we want to reject an update of a projection

Currently we do so that registering a projection is idempotent. Doing so makes updates more difficult though. The way to do it currently is `unregister_projection` and then `register_projection`

src/khepri_machine.erl
the-mikedavis: I added a function that wraps the `erlang:erase/1` call and some docs docs and put it next to `get_compiled_projection_tree/1` for extra clarity

src/khepri_machine.erl
dumbbell: I understand, thank you!

dumbbell (approved):
Thank you!

PR #256, opened 2024-04-03
Simplify `{error,timeout}` error type

the-mikedavis:
Previously this return value was '{error,{timeout,RaLeaderId}}'. Leader information is generally not useful to Khepri API consumers though since Khepri covers leader redirection. So this change removes the leader ID from the error type and returns '{error,timeout}' instead.


dumbbell (approved):
Thank you!

PR #257, opened 2024-05-15
khepri_pattern_tree: Fix conditions that were matching subtrees

dumbbell:
## Why

A pattern tree with `/parent/*` is supposed to match `/parent/child` but not `/parent/child/grandchild`.

Before this patch, this was the case because we didn't check if we reached the of the path before calling the folding function in `khepri_pattern_tree:fold1/6`.

## How

Now, we check the context and only call the fold function if we reached the end of the path.

Otherwise, we recurse without calling that fold function, checking a few more things depending on whether the condition could match grand children or not.

The testsuites are expanded with more testcases to cover these situations.

### DIFF
--- a/src/khepri_pattern_tree.erl
+++ b/src/khepri_pattern_tree.erl
@@ -149,9 +149,14 @@ update_payload(#pattern_node{payload = Payload} = PatternTree, UpdateFun) ->
 %% @see fold_acc().
 
 fold(PatternTree, Tree, Path, FoldFun, Acc) ->
-    Acc1 = fold_data(PatternTree, [], FoldFun, Acc),
-    Root = Tree#tree.root,
-    fold1(PatternTree, Root, Path, FoldFun, Acc1, []).
+    Path1 = khepri_path:realpath(Path),
+    case Path1 of
+        [] ->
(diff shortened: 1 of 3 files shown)


src/khepri_pattern_tree.erl
```diff
@@ -164,40 +169,69 @@ fold(PatternTree, Tree, Path, FoldFun, Acc) ->
...
+                                FoldFun, Acc0, ReversedPath1);
+                          true when AppliesToGrandchildren ->
+                              %% Same as above, but because this condition can
+                              %% be used on grand children, we two scenarios:
+                              %%   1. The condition may still match grand
+                              %%      children, so we keep it and continue down
+                              %%      the tree. This is `Acc1' below.
+                              %%   2. The condition won't match any grand
+                              %%      children (but it matched so far). We
+                              %%      evaluete the grand children with the
```
the-mikedavis, line 213:
```suggestion
                              %%      evaluate the grand children with the
```

the-mikedavis (commented):
Looks good, thanks for fixing this! I just found a tiny typo

the-mikedavis (approved):

PR #259, opened 2024-06-12
Restore projections when a snapshot is installed

the-mikedavis:
This is the companion PR to https://github.com/rabbitmq/ra/pull/448

The new `ra_machine:snapshot_installed/2` optional callback can be used to sync the Khepri tree and projection tables. Without syncing these, if a member is caught up by the leader with a snapshot, any changes between where the member was in the log and the new snapshot index are not reflected in projection tables. The second commit has a reproduction case and explanation of a bug that is possible if we do not sync the projection tables when a snapshot is installed.


the-mikedavis:
a808303 is not fully related to the `snapshot_installed` change, just something I noticed while setting up the new test case (in 24b2814)

dumbbell (changes requested):
Great patch, thank you!

Just two minor comments:

A typo in the commit message of the last commit: "the machine state maches the leader", "maches" -> "matches"

Still in the commit message, in the steps to reproduce the problem with RabbitMQ, could you please clarify in the entire paragraph when you talk about RabbitMQ? Because this is Khepri repository, RabbitMQ is kind of out of the context. For example, the first line of that section says "This fixes a bug reproducible in the server with the following reproduction steps:". It should be explicit that you are talking about the "RabbitMQ server". Likewise for the rest of the section when you talk about servers, nodes and so on.

dumbbell (approved):
Thanks!

PR #260, opened 2024-06-27
Improvements around queries

dumbbell:
Currently, queries suffer several issues or limitations:
* They are executed remotely on the leader. This is a problem if the remote node has an incompatible version of Erlang/OTP or the module hosting the function, or even lacks the function entirely.
* Local queries solve this problem but they may be executed against ouf-of-date data when compared to updates that were previously submitted.

This collection of patches aims at fixing this with the following changes:
* The default options of updates changes to use the `reply_from => local` option.
* The default query type is now `local` (possible thanks to the change above).
* A new `khepri:fence()` API is introduced.

See individual commits to learn more.


src/khepri.erl
```diff
@@ -288,36 +290,41 @@
...
+%% not sure that the given query function can be executed by a remote node
+%% (e.g. is the remote module available or compatible?). `local' is the default
+%% and recommended setting.</li>
+%% <li>`leader' to use {@link ra:leader_query/3}. The query is evaluated by the
+%% leading Ra server. It has more latency and may fail if the leader is
+%% currently unreachable, but it works on the latest data. The query can crash
+%% if the remote node lacks the query function's module or if the query
+%% function's remote copy is incompatible.</li>
+%% <li>`consistent' to use {@link ra:consistent_query/3}. The query is also
+%% evaluated by the leadin Ra server like `leader'. However, it is only
```
the-mikedavis, line 311:
```suggestion
%% evaluated by the leader Ra server like `leader'. However, it is only
```

or "leading"

src/khepri.erl
```diff
@@ -3403,6 +3410,77 @@ transaction(FunOrPath, Args, ReadWrite, Options)
...
+%% locally.
+%%
+%% This ensures that a subsequent query will see the result of synchronous and
+%% asynchronous updates.
+%%
+%% This can't work however if:
+%% <ul>
+%% <li>Asynchronous updates have a correlation ID, in which case the caller is
+%% responsible for waiting for the replies.</li>
+%% <li>The default `reply_from => local' command option is overriden by
```
the-mikedavis, line 3472:
```suggestion
%% <li>The default `reply_from => local' command option is overridden by
```

the-mikedavis (commented):
Just some small typos, otherwise this is looking good to me!

src/khepri.erl
dumbbell: Fixed!

src/khepri.erl
dumbbell: Done

dumbbell:
I renamed the existing testcase in `test/fence.erl` (too much copy-pasting :-) and added two more testcases.

the-mikedavis (approved):

dumbbell:
This pull request is ready for review again. @the-mikedavis, compared to the last time you looked at it, I reintroduced the `favor` option but limited the values to `low_latency` and `consistency`. Regardless of the value, the query is always local. The difference is in the use of the fence mechanism of not.

the-mikedavis:
I found a few typos in the commit messages:

* c5ea616 "loader" => "leader"
* c53ec8e "ais" => "is"
* b66fc4f "increaset" => "increase"
* b66fc4f "beraviors" => "behaviors"

the-mikedavis (approved):
Other than the typos this all looks good to me!

dumbbell:
Thank you! Thank was fast :-) I fixed the typos in the commit messages.

dumbbell:
The use of `reply_from => local` for all commands is useless as it is in this patch because we try to send commands directly to the leader if we know it... I believe this explains the transient failure in `cluster_SUITE` we se in CI.

the-mikedavis:
IIRC `reply_from` can also be `member => RaServerId`. We could pass the local node as that member

dumbbell:
> IIRC `reply_from` can also be `member => RaServerId`. We could pass the local node as that member

Good idea. I will restore the direct send to the leader if it is known and use the `reply_from => {member, LocalRaServerId}` option.

I'm in the process of removing the leader ID caching we do and use `ra_leaderboard:lookup_leader/1` directly.

dumbbell:
@the-mikedavis: I restored the send to the leader, but with `reply_from => {member, LocalRaServer}`. I also dropped the cached leader thing in favor of the use of `ra_leaderboard:lookup_leader/1`. I finally reorganized the commits a bit and reworded some commit messages. It's ready for review again and hopefully, it should be the last iteration :-)

the-mikedavis (approved):

PR #282, opened 2024-07-31
Replace `unregister_projection` command with `unregister_projections`

the-mikedavis:
The motivation for this in RabbitMQ is that we currently mistakenly register projections even when using mnesia as the metadata store in 3.13.x. It's a benign mistake - it shouldn't have much impact on those running 3.13.x - but it will be a potential headache for upgrading to 4.0.x since we also plan on reorganizing the tree layout (see https://github.com/rabbitmq/rabbitmq-server/pull/11225). With this change we will be able to use `khepri:unregister_all_projections/0` as we enable the `khepri_db` flag to remove those projections and add fresh definitions.

Implementation-wise the change is very straightforward: it's a new `#unregister_all_projections{}` command which calls `khepri_projection:delete/1` for each deleted projections and returns a map of the deleted projections (for `khepri_adv`). `khepri:unregister_all_projections/0,1,2` discards this map and returns the simple `ok | khepri:error()`.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -61,7 +61,7 @@
          transaction/5,
          register_trigger/5,
          register_projection/4,
-         unregister_projection/3]).
+         unregister_projections/3]).
 -export([get_keep_while_conds_state/2,
          get_projections_state/2]).
 
@@ -122,7 +122,7 @@
                    #register_trigger{} |
                    #ack_triggered{} |
                    #register_projection{} |
-                   #unregister_projection{} |
+                   #unregister_projections{} |
                    #dedup{} |
                    #dedup_ack{}.
 %% Commands specific to this Ra machine.
@@ -186,6 +186,10 @@
                              [khepri_projection:projection()]).
 %% A pattern tree that holds all registered projections in the machine's state.
 
+-type projection_map() :: #{khepri_projection:name() => khepri_path:pattern()}.
+%% A mapping between the names of projections and patterns to which each
+%% projection is registered.
+
 -export_type([common_ret/0,
               tx_ret/0,
               async_ret/0,
@@ -200,6 +204,7 @@
               props/0,
               triggered/0,
               projection_tree/0,
+              projection_map/0,
               command/0]).
 
 -define(HAS_TIME_LEFT(Timeout), (Timeout =:= infinity orelse Timeout > 0)).
@@ -593,23 +598,31 @@ register_projection(
                                    projection = Projection},
     process_command(StoreId, Command, Options).
 
--spec unregister_projection(StoreId, ProjectionName, Options) -> Ret when
+-spec unregister_projections(StoreId, Names, Options) -> Ret when
       StoreId :: khepri:store_id(),
-      ProjectionName :: khepri_projection:name(),
+      Names :: all | [khepri_projection:name()],
       Options :: khepri:command_options(),
-      Ret :: ok | khepri:error().
-%% @doc Unregisters a projection by name.
+      Ret :: khepri:ok(khepri_machine:projection_map()) | khepri:error().
+%% @doc Removes the given projections from the store.
+%%
+%% `Names' may either be a list of projection names to remove or the atom
+%% `all'. When `all' is passed, every projection in the store is removed.
 %%
 %% @param StoreId the name of the Ra cluster.
-%% @param ProjectionName the name of the projection to unregister.
+%% @param Names the names of projections to unregister or the atom `all' to
+%%        remove all projections.
 %% @param Options command options such as the command type.
 %%
-%% @returns `ok' if the projection existed and was unregistered, an `{error,
-%% Reason}' tuple otherwise.
+%% @returns `{ok, ProjectionMap}' if the command succeeds, `{error, Reason}'
+%% otherwise. The `ProjectionMap' is a map with projection names ({@link
+%% khepri_projection:name()} keys associated to the pattern to which each
+%% projection was registered.
 
-unregister_projection(StoreId, ProjectionName, Options)
-  when ?IS_KHEPRI_STORE_ID(StoreId) andalso is_atom(ProjectionName) ->
-    Command = #unregister_projection{name = ProjectionName},
+unregister_projections(StoreId, Names, Options)
+  when ?IS_KHEPRI_STORE_ID(StoreId) andalso
+       (Names =:= all orelse is_list(Names)) andalso
+       is_map(Options) ->
+    Command = #unregister_projections{names = Names},
     process_command(StoreId, Command, Options).
 
 -spec ack_triggers_execution(StoreId, TriggeredStoredProcs) ->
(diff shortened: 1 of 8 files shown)


the-mikedavis:
There was a proptest failure in the last CI run: https://github.com/rabbitmq/khepri/actions/runs/10187161148/job/28180580513?pr=282

I don't think it's related to these changes but I'm not very familiar with proptest https://github.com/rabbitmq/khepri/actions/runs/10187161148/job/28180580513?pr=282

dumbbell:
Have you considered modifying the `unregister_projection` command instead of adding a new one?

I don’t know if it makes sense or could be useful, but the existing command could take a regex:
```erlang
#unregister_projection{name = ".*"}
```

Just wondering if we can be more future proof here.

src/khepri_pattern_tree.erl
```diff
@@ -255,6 +257,36 @@ fold_data(PatternTreeNode, CurrentPath, FoldFun, Acc) ->
             FoldFun(CurrentPath, Payload, Acc)
     end.
 
+-spec fold(PatternTree, Fun, Acc) -> Ret when
+      PatternTree :: khepri_pattern_tree:tree(Payload),
+      Fun :: fold_fun(Payload),
+      Acc :: fold_acc(),
+      Ret :: fold_acc(),
+      Payload :: payload().
```
dumbbell, line 265: Style commen: could you please move this function before `fold_matching()`? Like, the more generic one before the more specialized one. And for the alphabetical order.

src/khepri_pattern_tree.erl
```diff
@@ -255,6 +257,36 @@ fold_data(PatternTreeNode, CurrentPath, FoldFun, Acc) ->
...
+%% accumulator to the fold function.
+%%
+%% Unlike {@link fold_matching/5} all payloads are passed to the fold function.
+
+fold(#pattern_node{} = PatternTree, Fun, Acc) ->
+    fold(PatternTree, Fun, [], Acc).
+
+fold(
+  #pattern_node{payload = Payload, child_nodes = ChildNodes},
+  Fun, ReversedPath, Acc0) ->
```
dumbbell, line 276: Style comment: could you please name the argument `Acc` and then increment the suffix in the function body? I like that the arguments have the "normal" name, the same as the type spec and the doc. I know, this is the opposite of Ra where the normal name is used for the returned value :-)

test/projections.erl
```diff
@@ -671,3 +671,76 @@ unregister_projection_test_() ->
...
+unregister_all_projections_test_() ->
+    ProjectFun = fun(Path, Payload) -> {Path, Payload} end,
+    PathPattern = [stock, wood, <<"oak">>],
+    Data = 100,
+    ProjectionName1 = projection_1,
+    ProjectionName2 = projection_2,
+    {setup,
+     fun() -> test_ra_server_helpers:setup(?FUNCTION_NAME) end,
+     fun(Priv) -> test_ra_server_helpers:cleanup(Priv) end,
+     [{inorder,
```
dumbbell, line 684: Could you please test that `unregister_all_projections()` works with no projections at all?

src/khepri_machine.erl
```diff
@@ -1303,6 +1328,22 @@ apply(
...
     Ret = {State1, Reply},
     post_apply(Ret, Meta);
+apply(Meta, #unregister_all_projections{}, State) ->
+    ProjectionTree = get_projections(State),
+    Projections = khepri_pattern_tree:fold(
+                   ProjectionTree,
+                   fun(PathPattern, Projections, Acc) ->
+                           lists:foldl(
+                             fun(Projection, Acc1) ->
+                                     khepri_projection:delete(Projection),
```
dumbbell, line 1338: I’m wondering, should we protect the code against an exception from ETS and return an error? It could take down the Ra server otherwise.

dumbbell (changes requested):

src/khepri_machine.erl
the-mikedavis: We should add a `catch` in `khepri_projection:delete/1` but the existing callers (and this call here) should ignore the error I think. If we get an error that means that the table doesn't exist which is the state we want it in. I don't think that should be possible (at this point the table should exist) but it would be nice to protect against something unexpected killing the server here

the-mikedavis:
We could extend the `unregister_projection` command a little so we have special cases:

```erlang
#unregister_projection{name = {name, khepri_projection:name()} | all}
```

And in the future we could add a regex on a list of names to unregister. What do you think? I'd like to avoid the regex if there isn't a clear use-case for it. I think the most common use-cases would be removing all projections or a list of names you know ahead of time (`[khepri_projection:name()]`)

dumbbell:
I like your idea to accept a list of names. Then, what about replacing the existing command by:

```erlang
#unregister_projections{names = [khepri_projection:name()] | all}
```

(so, plural for the command name and argument)

the-mikedavis:
I've replaced `khepri:unregister_projection/1,2,3` with `khepri:unregister_projections/1,2,3` that accepts either a list of projection names or the `all` atom. The old behavior can be emulated with `khepri:unregister_projections([Name]).

It behaves idempotently, so instead of returning `{error, not_found}` we return `ok` when the command succeeds. There is a new `khepri_adv:unregister_projections/1,2,3` that returns a map of removed projections in the shape of `#{khepri_projection:name() => khepri_path:native_pattern()}` which can be used to detect when a projection was actually removed or not.

src/khepri_machine.erl
```diff
@@ -1268,39 +1281,41 @@ apply(
...
-                      {true, Projections1}
-              end
-      end),
-    Reply = case ProjectionTree1 of
-                ProjectionTree ->
-                    Info = #{name => ProjectionName},
-                    Reason = ?khepri_error(projection_not_found, Info),
-                    {error, Reason};
-                _ ->
-                    clear_compiled_projection_tree(),
```
dumbbell, line 1300: Don’t we need this call to `clear_compiled_projection_tree()`?

the-mikedavis: Whoops, good catch!

test/cluster_SUITE.erl
```diff
@@ -1520,13 +1520,11 @@ can_use_default_store_on_single_node(_Config) ->
...
+    ?assertEqual(ok, khepri:unregister_projections([ProjectionName1])),
     ?assertEqual(
-       {error, ?khepri_error(
-                 projection_not_found,
-                 #{name => ProjectionName1})},
-       khepri:unregister_projection(ProjectionName1)),
-    ?assertEqual(ok, khepri:unregister_projection(ProjectionName2, #{})),
+       {ok, #{}},
+       khepri_adv:unregister_projections([ProjectionName1])),
+    ?assertEqual(ok, khepri:unregister_projections([ProjectionName2], #{})),
```
dumbbell, line 1527: Could you please add a call to `khepri_adv:unregister_projections/2` as well?

Mostly for test coverage.

src/khepri_projection.erl
```diff
@@ -257,9 +263,23 @@ init(#khepri_projection{name = Name, ets_options = EtsOptions}) ->
...
 delete(#khepri_projection{name = Name}) ->
-    _ = ets:delete(Name),
-    ok.
+    try
+        ets:delete(Name),
+        ok
+    catch
+        error:badarg:_Stacktrace ->
+            {error, does_not_exist}
+    end.
```
dumbbell, line 282: Could you please add an `unregister_projections()` test with an ETS table removed behind the scene and check that the Ra server doesn’t crash?

dumbbell (changes requested):
The patch looks good to me.

I have two last comments about new test cases.

dumbbell (approved):
Thank you!

PR #283, opened 2024-08-13
Improve the fence mechanism performances

dumbbell:
This pull request is a collection of patches that improves the performances of the fence mechanism.

It uses `ra:key_metrics/2` instead of `ra:member_overview/2` and avoids the preliminary query that we do to serialize events if it is not needed. See individual commits for more details.

This pull request is blocked because it depends on a new release of Ra that contains rabbitmq/ra#461. Also, to fully fix the performance problem introduced with the fence mechanism, we also need that that future release of Ra contains rabbitmq/ra#462.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -865,8 +865,12 @@ do_process_sync_command(StoreId, Command, Options) ->
     CommandOptions = #{timeout => Timeout, reply_from => ReplyFrom},
     T0 = khepri_utils:start_timeout_window(Timeout),
     Dest = case ra_leaderboard:lookup_leader(StoreId) of
-               LeaderId when LeaderId =/= undefined -> LeaderId;
-               undefined                            -> RaServer
+               LeaderId when LeaderId =/= undefined ->
+                   sending_command_remotely(StoreId),
+                   LeaderId;
+               undefined ->
+                   sending_sync_command_locally(StoreId),
+                   RaServer
            end,
     case ra:process_command(Dest, Command, CommandOptions) of
         {ok, Ret, _LeaderId} ->
@@ -903,15 +907,18 @@ process_async_command(
   StoreId, Command, ?DEFAULT_RA_COMMAND_CORRELATION = Correlation, Priority) ->
     ThisNode = node(),
     RaServer = khepri_cluster:node_to_member(StoreId, ThisNode),
+    sending_async_command_locally(StoreId),
     ra:pipeline_command(RaServer, Command, Correlation, Priority);
 process_async_command(
   StoreId, Command, Correlation, Priority) ->
     case ra_leaderboard:lookup_leader(StoreId) of
         LeaderId when LeaderId =/= undefined ->
+            sending_command_remotely(StoreId),
             ra:pipeline_command(LeaderId, Command, Correlation, Priority);
         undefined ->
             ThisNode = node(),
             RaServer = khepri_cluster:node_to_member(StoreId, ThisNode),
+            sending_async_command_locally(StoreId),
             ra:pipeline_command(RaServer, Command, Correlation, Priority)
     end.
 
@@ -983,6 +990,7 @@ process_query(StoreId, QueryFun, Options) ->
     end.
 
 process_query1(StoreId, QueryFun, Options) ->
+    sending_query_locally(StoreId),
     LocalServerId = {StoreId, node()},
     case ra:local_query(LocalServerId, QueryFun, Options) of
         {ok, {_RaIdxTerm, Ret}, _NewLeaderId} ->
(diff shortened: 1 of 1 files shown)


src/khepri_machine.erl
```diff
@@ -1020,63 +1028,68 @@ add_applied_condition1(StoreId, Options, Timeout) ->
...
-            Condition = {applied, {LastIndex, CurrentTerm}},
-            Options1 = Options#{condition => Condition,
-                                timeout => NewTimeout},
-            {ok, Options1};
-        {ok, _Overview, NewLeaderId} ->
-            NewTimeout = khepri_utils:end_timeout_window(Timeout, T0),
-            add_applied_condition3(StoreId, Options, NewLeaderId, NewTimeout);
-        {timeout, _LeaderId} ->
+    try
+        case ra:key_metrics(LeaderId, Timeout) of
```
the-mikedavis, line 1065: I believe the syntax for try + case can be combined:

```erl
try ra:key_metrics(LeaderId, Timeout) of
    #{last_index := LastIndex, ..} ->
        %% ...
    _ ->
        %% ...
catch
    error:{erpc, timeout} ->
        %% ...
end
```

It might cut down on nesting a bit. (Not something I feel strongly about though if you prefer the separate try + case)

src/khepri_machine.erl
```diff
@@ -1096,6 +1109,43 @@ get_timeout(_)                     -> khepri_app:get_default_timeout().
...
+last_message_was_sync(StoreId) ->
+    Key = ?LAST_MSG_WAS_SYNC_KEY(StoreId),
+    _ = erlang:put(Key, true),
+    ok.
+
+-spec last_message_was_async(StoreId) -> ok when
+      StoreId :: khepri:store_id().
+%% @doc Remenbers that the calling process just made an asynchronous command.
+
+last_message_was_async(StoreId) ->
```
the-mikedavis, line 1129: Since we also use this function for the case where we don't know the leader or haven't sent a command yet I think it might be clearer to name this `last_message_was_not_sync/1`. To me `last_message_was_async` implies that we did send an async message which may not be the case - just that the last message wasn't sent synchronously

src/khepri_machine.erl
dumbbell: Good idea, I will do that.

src/khepri_machine.erl
dumbbell: I’m not satisfied with the names either.

The idea is to distinguish a sync command or a query, from an async command, an absence of any message sent to the local Ra server or a message sent to another Ra server.

Perhaps we should distinguish each of these situation:
```erlang
%% Sets the internal flag to true.
sending_query_locally(StoreId),
sending_sync_command_locally(StoreId),

%% Clears the internal flag.
sending_async_command_locally(StoreId),
sending_command_remotely(StoreId).
```

Then, the internal flag  could be named `skip_fence_preliminary_query` and the `was_last_message_sync()` renamed to `can_skip_fence_preliminary_query(StoreId)`.

the-mikedavis: That sounds good to me :+1:

Even if it's a bit more verbose I think it's clearer about its intentions

dumbbell: Cool, I will proceed with these names then. Thanks!

dumbbell:
This pull request is unblocked by the release of Ra 2.13.6.

src/khepri_machine.erl
```diff
@@ -1096,6 +1110,71 @@ get_timeout(_)                     -> khepri_app:get_default_timeout().
...
+-spec can_skip_fence_preliminary_query(StoreId) -> LastMsgWasSync when
+      StoreId :: khepri:store_id(),
+      LastMsgWasSync :: boolean().
+%% @doc Indicates if the calling process sent a synchronous command or a query
+%% before this call.
+%%
+%% @returns `true' if the calling process sent a synchrorous command or a
+%% query to the given store before this call, `false' if the calling process
+%% never sent anything to the given store, or if the last message was an
+%% asynchrorous command, th
```
the-mikedavis, line 1172: Looks like this line got cut off here

the-mikedavis (commented):
Just a tiny comment about docs, otherwise looks good

dumbbell:
For the record, my local testing shows that consistent queries are about 50% slower thar low latency queries.

Before the whole work in #260 and this pull request, consistent queries were slower by a single digit percent.

the-mikedavis (approved):

the-mikedavis:
I think consistent queries being slower than low latency is not very concerning for the sake of RabbitMQ: we only use consistent queries very rarely

dumbbell:
Indeed. Also sync writes and sync transactions now wait for the command to be applied locally before returning. Therefore, consistent queries should be rarely useful.

PR #286, opened 2024-08-16
khepri_machine: Restore `#unregister_projection{}` for backward-compatibility

dumbbell:
## Why

Even though this version of Khepri never emits `#unregister_projection{}`, existing WAL files may still contain that command.

## How

To not break these existing deployment, `apply/3` accepts `#unregister_projection{}` again. It converts it to the new `#unregister_projections{}` command and recurses.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -127,6 +127,15 @@
                    #dedup_ack{}.
 %% Commands specific to this Ra machine.
 
+-type old_command() :: #unregister_projection{}.
+%% Old commands that are still accepted by the Ra machine but never created.
+%%
+%% Even though Khepri no longer creates these commands, they may still be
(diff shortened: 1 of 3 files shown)


the-mikedavis (commented):
Codewise :+1:

For the commit description and the comment in `apply/3` though I think it's more general to say that the command might be "in existing Raft logs". The WAL is a temporary place to store commands as a performance optimization and the command might also be in segment files. But both are part of the Raft log.

dumbbell:
Good point @the-mikedavis, I improved the wording.

the-mikedavis (approved):

dumbbell:
I added a test case that just makes sure the machine code won't crash. It doesn't verify if the projections are actually unregistered.

PR #287, opened 2024-08-22
Fixes to `khepri_cluster` member queries

dumbbell:
This pull request contains a set of patches to fix the functions to query Ra members and nodes in `khepri_cluster`:
* The return value in case of a timeout was not aligned with the type spec and inconsistent with what other functions do.
* The timeout argument may not have been honored in the case of a `noproc`.
* The query was not retried after a `noconnection` error and a few other related errors.

### DIFF
--- a/src/khepri_cluster.erl
+++ b/src/khepri_cluster.erl
@@ -1236,6 +1236,16 @@ do_query_members(StoreId, RaServer, QueryType, Timeout) ->
                        [StoreId]),
                     Error
             end;
+        {error, Reason}
+          when ?HAS_TIME_LEFT(Timeout) andalso
+               (Reason == noconnection orelse
+                Reason == nodedown orelse
+                Reason == shutdown) ->
+            NewTimeout0 = khepri_utils:end_timeout_window(Timeout, T0),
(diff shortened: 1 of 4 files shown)


src/khepri_cluster.erl
```diff
@@ -1235,6 +1236,21 @@ do_query_members(StoreId, RaServer, QueryType, Timeout) ->
...
                     Error
             end;
+        {error, Reason}
+          when ?HAS_TIME_LEFT(Timeout) andalso
+               (Reason == noconnection orelse
+                Reason == nodedown orelse
+                Reason == shutdown) ->
+            NewTimeout0 = khepri_utils:end_timeout_window(Timeout, T0),
+            NewTimeout = khepri_utils:sleep(
+                           ?NOPROC_RETRY_INTERVAL, NewTimeout0),
```
the-mikedavis, line 1246: Not totally related to this change but should we rename this constant to something like `TRANSIENT_ERROR_RETRY_INTERVAL` now that we use it for other error reasons?

dumbbell: You are right. I can prepare another pull request.

the-mikedavis (approved):

PR #289, opened 2024-08-23
khepri_cluster: Add `wait_for_leader/{0,1,2}`

dumbbell:
## Why

This new public API is useful if you want to be sure a clustered store is ready before issueing writes and queries.

Note that there are obviously no guaranties that the Raft quorum will be lost just after this call.

## How

The implementation is very close to `members/{0,1,2}`. The difference are:
* how errors are handled
* the return value

### DIFF
--- a/src/khepri_cluster.erl
+++ b/src/khepri_cluster.erl
@@ -115,6 +115,7 @@
          locally_known_nodes/0,
          locally_known_nodes/1,
          locally_known_nodes/2,
+         wait_for_leader/0, wait_for_leader/1, wait_for_leader/2,
          get_default_ra_system_or_data_dir/0,
          get_default_store_id/0,
          get_store_ids/0,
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):

the-mikedavis (approved):

dumbbell:
I reverted this back to a draft: there is a failure when I test this with RabbitMQ I don’t fully understand. In a testcase that resets a RabbitMQ node that is in minority, this `khepri_cluster:wait_for_leader()` function times out while `rabbit_khepri:wait_for_leader()` doesn’t. The only diffirence is the loop that the latter does to retry.

I need to undestand why the retry convinces a leader to appear :-)

dumbbell:
The problem was in the modifications I made to the RabbitMQ code when I adapted it to use this new API.

PR #290, opened 2024-08-23
Update all projections on snapshot installation

the-mikedavis:
When a snapshot is installed the machine state is "swapped" from the state before the snapshot to the one contained in the snapshot. This swap jumps over potentially many commands in the log which we would have used to trigger projections. When we install a snapshot we need to update the projection state but it's not as straightforward as when "restoring" projections (on machine recovery for example) because the projection tables will already exist and potentially contain records.

When we detect that a snapshot has been installed we need to diff the old and new states to find which projections have changed themselves and also update their associated projection tables.

This requires a change in Ra to expose the old machine state and version in the `snapshot_installed` callback: https://github.com/rabbitmq/ra/pull/467


src/khepri_machine.erl
```diff
@@ -2322,3 +2340,126 @@ convert_state(State, 0, 1) ->
...
+
+    CommonProjections = sets:intersection(OldProjections, NewProjections),
+    DeletedProjections = sets:subtract(OldProjections, CommonProjections),
+    CreatedProjections = sets:subtract(NewProjections, CommonProjections),
+
+    %% Tear down any projections which were unregistered.
+    lists:foreach(
+      fun({_Pattern, Projection}) ->
+              _ = khepri_projection:delete(Projection)
+      end, sets:to_list(DeletedProjections)),
```
dumbbell, line 2375: Could the use of `sets:fold/3` be more efficient?

src/khepri_machine.erl
```diff
@@ -2322,3 +2340,126 @@ convert_state(State, 0, 1) ->
...
+                        sets:add_element(Entry, Acc1)
+                end, Acc, Projections)
+      end, sets:new([{version, 2}])).
+
+update_projection(Pattern, Projection, OldTree, NewTree) ->
+    TreeOptions = #{props_to_return => ?PROJECTION_PROPS_TO_RETURN,
+                    include_root_props => true},
+    case khepri_tree:find_matching_nodes(OldTree, Pattern, TreeOptions) of
+        {ok, OldMatchingNodes} ->
+            case khepri_tree:find_matching_nodes(NewTree, Pattern, TreeOptions) of
```
dumbbell, line 2421: This line is too long. Could you please assign the result of `khepri_tree:find_matching_nodes(NewTree, Pattern, TreeOptions)` to a variable and use the variable in the `case`?

test/cluster_SUITE.erl
```diff
@@ -1916,44 +1917,78 @@ projections_are_updated_when_a_snapshot_is_installed(Config) ->
...
       "Stop cluster member ~s (quorum is maintained)", [Node1]),
     ok = rpc:call(Node1, khepri, stop, [StoreId]),
 
-    ?assertMatch(
-      {ok, #{log := #{snapshot_index := undefined}}, _},
-      ra:member_overview(khepri_cluster:node_to_member(StoreId, Node3))),
-
-    ct:pal("Submit enough commands to trigger a snapshot"),
+    ct:pal("Modify paths which are watched by projections"),
     ct:pal("- set key1:value1v2"),
```
dumbbell, line 1967: This message isn’t close to the line it describes anymore.

src/khepri_machine.erl
the-mikedavis: I expect that we shouldn't have so many projections that it becomes noticeable but it does seem needless to me to convert to a list just for foreach. I switched to `sets:fold/3`.

dumbbell: You are right, it will probably make little difference.

src/khepri_machine.erl
```diff
@@ -1253,6 +1253,12 @@ handle_aux(
                Projection <- Projections]
       end),
     {no_reply, AuxState, LogState};
+handle_aux(
+  _RaState, cast, {update_projections, OldState}, AuxState, LogState,
+  NewState) ->
+    ok = update_projections(OldState, NewState),
```
kjnilsson, line 1259: could this work not be done in `snapshot_installed/4` instead of delegating to aux?

src/khepri_machine.erl
```diff
@@ -1253,6 +1253,12 @@ handle_aux(
                Projection <- Projections]
       end),
     {no_reply, AuxState, LogState};
+handle_aux(
```
kjnilsson, line 1256: why not make use of the screen estate and put some of the arguments on this line? with this formatting the function head takes up 3 lines when it could just use 2 (and maintain left to right reading)

src/khepri_machine.erl
the-mikedavis: yeah good point, I've removed the `update_projections` effect and moved the handler for that inline here

src/khepri_machine.erl
the-mikedavis: I don't have an opinion either way, this was just to fit the existing indentation in the project

dumbbell: Yeah, that’s me who is not a fan of variable indentation where you end up with:
```erlang
function(Arg1,
         Arg2)
```

If all arguments can’t fit on one line, I put all of them on the next line:
```erlang
function(
  Arg1,
  Arg2)
```

By "variable indentation", I mean that the column on which the arguments start would depend on the length of the function name. In almost all cases, they would start more on the right than the function body:
```erlang
function(Arg1,
         Arg2) ->
    do().
```
Instead of keeping some kind of hierarchy where the function body is more indented than the function clause:
```erlang
function(
  Arg1,
  Arg2) ->
    do().
```

dumbbell (commented):
The patch looks good to me. Great work!

Could you just please squash the commit adding `khepri_pattern_tree:fold()` into the commit that uses it? This addition goes with it and is only tested with it. What do you think?

the-mikedavis:
I squashed the `fold/3` change into the main change (last commit) and added a small blurb to that commit message about `fold/3` :+1:

dumbbell (approved):

PR #291, opened 2024-08-24
khepri_tx_adv: Prefer list prepending to build side effect lists

the-mikedavis:
This is an optimization for functions in the transaction APIs. Specifically this should be very noticeable when any command from the `khepri_tx` or `khepri_tx_adv` API (delete, put, etc.) is used with a pattern that affects very many nodes (10,000 or more). The gist is that we can get surprisingly nice speed ups for some functions by building our side effect lists with `[Effect | Effects]` rather than `++/2`.

For background: I have WIP changes to simplify the exchange deletion part of vhost deletion in RabbitMQ by using a transaction ([commit](https://github.com/rabbitmq/rabbitmq-server/commit/2763111bf6e9)). It deletes all exchanges and bindings in a single command without adding/removing runtime parameters and returns the combined `rabbit_bindings:deletions()` dict. It also changes the way we delete bindings in a transaction to use `khepri_tx_adv:delete_many/1` (which returns a map of the deletions) rather than `khepri_tx:get_many/1` plus `khepri_tx:delete/1` for each binding. Against a stress test of deleting a vhost with 50,000 exchanges with no bindings or queues, I expected the deletion to be almost instant but `rabbitmqctl delete_vhost myvhost` took around 500s.

A flamegraph points to the BIF behind `erlang:'++'/2` (and allocation functions associated with list appending) dominating the time taken to execute a `khepri_tx_adv:delete_many/1` call in this block: https://github.com/rabbitmq/khepri/blob/be71afe3eff3cbb6b055372ced587aa63d0f4276/src/khepri_tx_adv.erl#L997-L1006

<details><summary>Flamegraph...</summary>

Zoomed in on this function in particular:

flamegraph

[perf.data.zip](https://github.com/user-attachments/files/16736792/perf.data.zip)

---

</details>

That is most expensive when deleting bindings which is odd since this test vhost only has these 50,000 exchanges - no queues or bindings or anything else.

`List ++ []` is not optimized in Erlang/OTP (see https://github.com/erlang/otp/pull/8743): `erlang:'++'/2` clones the left-hand-side unless it is `[]` which spends time and also introduces a lot of garbage. The reason that the reproduction case is slow is that we basically call `List ++ []` twice for the non-existent bindings of each exchange we delete which is O(n^2) behavior (`n` being the number of exchanges) since cloning the list takes O(n) time and memory. List prepending is O(1) but even better we don't need to pay any cost when we don't end up prepending a new element.

Regardless of whether `List ++ []` is optimized in Erlang/OTP we should switch to prepending: a large transaction that issues many individual commands will also cause many appends and the right-hand side of that append might be non-empty.

<details><summary>Reproduction steps in RabbitMQ...</summary>

With the RabbitMQ repository checked out at branch `md/simplify-vhost-exchange-deletion` or the linked commit above:

1. Start a single-node broker with `make run-broker`.
1. Add a blank vhost: `sbin/rabbitmqctl add_vhost myvhost`.
1. Enable Khepri: `sbin/rabbitmqctl enable_feature_flag --experimental khepri_db`.
1. Create many exchanges. For example `[rabbit_exchange:declare({resource,<<"myvhost">>,exchange,erlang:list_to_binary(lists:flatten(io_lib:format("e~b", [N])))},topic,true,false,false,[],<<"rmq-internal">>) || N <- lists:seq(1,50_000)]` in the the shell or an eval command.
1. Delete the vhost: `time sbin/rabbitmqctl delete_vhost myvhost --timeout 10000`.

When the server is running with this branch the deletion takes around 12s. Before this change the deletion takes around 500s.

---

</details>

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -89,8 +89,8 @@
          split_query_options/1,
          split_command_options/1,
          split_put_options/1,
-         insert_or_update_node/5,
-         delete_matching_nodes/3,
+         insert_or_update_node/6,
+         delete_matching_nodes/4,
          handle_tx_exception/1,
          process_query/3,
          process_command/3]).
@@ -1295,13 +1295,13 @@ apply(
   State) ->
     {TreeOptions, PutOptions} = split_put_options(TreeAndPutOptions),
     Ret = insert_or_update_node(
-            State, PathPattern, Payload, PutOptions, TreeOptions),
+            State, PathPattern, Payload, PutOptions, TreeOptions, []),
     post_apply(Ret, Meta);
 apply(
   Meta,
   #delete{path = PathPattern, options = TreeOptions},
   State) ->
-    Ret = delete_matching_nodes(State, PathPattern, TreeOptions),
+    Ret = delete_matching_nodes(State, PathPattern, TreeOptions, []),
     post_apply(Ret, Meta);
 apply(
   Meta,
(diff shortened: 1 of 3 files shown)


michaelklishin (approved):

src/khepri_machine.erl
```diff
@@ -1709,54 +1709,58 @@ failed_to_locate_sproc(Reason) ->
     khepri_tx:abort(Reason).
 
 -spec insert_or_update_node(
-    State, PathPattern, Payload, PutOptions, TreeOptions) -> Ret when
+    State, PathPattern, Payload, PutOptions, TreeOptions, SideEffects) -> Ret when
```
dumbbell, line 1712: Could you please wrap that long line after the `->`?

src/khepri_machine.erl
```diff
@@ -1709,54 +1709,58 @@ failed_to_locate_sproc(Reason) ->
...
       PutOptions :: khepri:put_options(),
       TreeOptions :: khepri:tree_options(),
-      Ret :: {State, Result} | {State, Result, ra_machine:effects()},
+      SideEffects :: ra_machine:effects(),
+      Ret :: {State, Result, ra_machine:effects()},
       Result :: khepri_machine:common_ret().
 %% @private
 
-insert_or_update_node(State, PathPattern, Payload, PutOptions, TreeOptions) ->
+insert_or_update_node(State, PathPattern, Payload, PutOptions, TreeOptions, SideEffects) ->
```
dumbbell, line 1723: Likewise, could you please wrap the line after the `(`?

src/khepri_machine.erl
```diff
@@ -1709,54 +1709,58 @@ failed_to_locate_sproc(Reason) ->
...
     Ret1 = khepri_tree:insert_or_update_node(
              Tree, PathPattern, Payload, PutOptions, TreeOptions),
     case Ret1 of
         {ok, Tree1, AppliedChanges, Ret2} ->
             State1 = set_tree(State, Tree1),
-            {State2, SideEffects} = create_tree_change_side_effects(
-                                      State, State1, Ret2, AppliedChanges),
-            {State2, {ok, Ret2}, SideEffects};
+            {State2, SideEffects1} = add_tree_change_side_effects(
+                                       State, State1, Ret2, AppliedChanges, SideEffects),
```
dumbbell, line 1731: Likewise, could you please wrap the line after `AppliedChanges,`?

dumbbell (commented):
Great find!

I only have cosmetic comments.

the-mikedavis:
Forgot to clean up and makes sure things were wrapped properly, whoops!

I've also added a test case to assert that we handle the effects in the proper order.

src/khepri_machine.erl
```diff
@@ -1766,24 +1773,23 @@ create_tree_change_side_effects(
...
                 end, Changes0),
-    ProjectionEffects = create_projection_side_effects(
-                          InitialState, NewState, Changes),
-    {NewState1, TriggerEffects} = create_trigger_side_effects(
-                                    InitialState, NewState, Changes),
-    {NewState1, ProjectionEffects ++ TriggerEffects}.
+    SideEffects1 = add_projection_side_effects(
+                     InitialState, NewState, Changes, SideEffects),
+    add_trigger_side_effects(
+      InitialState, NewState, Changes, SideEffects1).
```
dumbbell, line 1779: The trigger-related side effects will appear before the projection-related ones now, won’t they?

Perpaps we should reverse the lines so that trigger-related side effects continue to appear after, in case a trigger function does a read from a projection. What do you think?

the-mikedavis: Hmm yeah, let's try to avoid changing the order effects are processed with this branch. I will also add some `lists:reverse/1` calls in `khepri_machine:apply/3` where these effects are handled so that the order is consistent between transactions and regular commands

dumbbell (approved):

PR #293, opened 2024-09-10
Handle deleted projections tables while triggering projections

the-mikedavis:
The test case in this change shows an example that can cause an error log. One client of Khepri sends an `unregister_projections` command while another sends many changes which affect a projection which is being unregistered. Using async commands makes the error much more likely to occur.

Ra applies committed commands in a batch and then handles all effects created by the batch. In this scenario the `unregister_projections` command may end up being applied by Ra in a batch with the async commands and the resulting effects list will contain aux effects to trigger projections (casued by the async commands) and an aux effect to unregister the projection. Some trigger aux effects may come later in the list than the `unregister_projections` effect and so their projection table will be deleted. In this case we want to perform a no-op for the trigger effect.

We already no-op these effects because we wrap ETS calls in `try` expressions. This change avoids an error log caused by ETS calls attempting to insert or delete records in an `undefined` table.

<details><summary>Example error log...</summary>

This is from the commontest output of the added test case before the fix in `khepri_projection.erl`:

```
2024-09-10T09:48:20.686591-04:00 <0.173.0>: Failed to insert record into ETS table for projection: cluster_SUITE
  Record:
    {[key],116}
  Crash:
    exception error: bad argument
     in function  ets:insert/2
        called as ets:insert(undefined,{[key],116})
        *** argument 1: the table identifier does not refer to an existing ETS table
     in call from khepri_projection:try_ets_insert/3 (/nix/persist/home/michael/src/rabbitmq/khepri/src/khepri_projection.erl, line 446)
     in call from khepri_projection:trigger_simple_projection/6 (/nix/persist/home/michael/src/rabbitmq/khepri/src/khepri_projection.erl, line 405)
     in call from khepri_machine:handle_aux/6 (/nix/persist/home/michael/src/rabbitmq/khepri/src/khepri_machine.erl, line 1242)
     in call from ra_server:handle_aux/4 (/nix/persist/home/michael/src/rabbitmq/khepri/_build/default/lib/ra/src/ra_server.erl, line 1632)
     in call from ra_server_proc:handle_effect/5 (/nix/persist/home/michael/src/rabbitmq/khepri/_build/default/lib/ra/src/ra_server_proc.erl, line 1394)
     in call from lists:foldl_1/3 (lists.erl, line 1599)
     in call from ra_server_proc:handle_effects/5 (/nix/persist/home/michael/src/rabbitmq/khepri/_build/default/lib/ra/src/ra_server_proc.erl, line 1301)
```

### DIFF
--- a/src/khepri_projection.erl
+++ b/src/khepri_projection.erl
@@ -303,18 +303,28 @@ delete(#khepri_projection{name = Name}) ->
 trigger(
   #khepri_projection{name = Name, projection_fun = ProjectionFun},
   Path, OldProps, NewProps) ->
-    Table = ets:whereis(Name),
-    case ProjectionFun of
-        copy ->
-            trigger_copy_projection(Name, Table, OldProps, NewProps);
-        StandaloneFun ->
-            case ?HORUS_STANDALONE_FUN_ARITY(StandaloneFun) of
-                2 ->
-                    trigger_simple_projection(
-                      Table, Name, StandaloneFun, Path, OldProps, NewProps);
-                4 ->
-                    trigger_extended_projection(
-                      Table, Name, StandaloneFun, Path, OldProps, NewProps)
+    case ets:whereis(Name) of
+        undefined ->
+            %% A table might have been deleted by an `unregister_projections'
+            %% effect in between when a `trigger_projection' effect is created
+            %% and when it is handled in `khepri_machine:handle_aux/6`. In this
+            %% case we should no-op the trigger effect.
+            ok;
+        Table ->
+            case ProjectionFun of
+                copy ->
+                    trigger_copy_projection(Name, Table, OldProps, NewProps);
+                StandaloneFun ->
+                    case ?HORUS_STANDALONE_FUN_ARITY(StandaloneFun) of
+                        2 ->
+                            trigger_simple_projection(
+                              Table, Name, StandaloneFun, Path,
+                              OldProps, NewProps);
+                        4 ->
+                            trigger_extended_projection(
+                              Table, Name, StandaloneFun, Path,
+                              OldProps, NewProps)
+                    end
             end
     end.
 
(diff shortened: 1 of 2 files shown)


src/khepri_projection.erl
```diff
@@ -303,18 +303,28 @@ delete(#khepri_projection{name = Name}) ->
...
-            trigger_copy_projection(Name, Table, OldProps, NewProps);
-        StandaloneFun ->
-            case ?HORUS_STANDALONE_FUN_ARITY(StandaloneFun) of
-                2 ->
-                    trigger_simple_projection(
-                      Table, Name, StandaloneFun, Path, OldProps, NewProps);
-                4 ->
-                    trigger_extended_projection(
-                      Table, Name, StandaloneFun, Path, OldProps, NewProps)
+    case ets:whereis(Name) of
```
dumbbell, line 306: I have a question first, mostly for my curiosity: why did we resolve the table name first instead of passing it to ETS?

dumbbell: And another question about the fix: which process executes this code?

In other words, can aux effects be executed concurrently?

I’m wondering if the ETS table can be deleted between that `ets:whereis/1` and the actual `ets:{insert,delete}()`. If aux effects are always executed sequentially for now, I think I would be more confortable if the "table doesn’t exist" error was caught by the `catch` later, to make the check atomic. This would be more future-proof as well if the design of aux effects changes some day.

Something like:
```erlang
try
    ets:insert(...)
catch
    Class:Reason:Stacktrace ->
        case ets:whereis(Name) of
            undefined ->
                %% Table doesn't exist
            _ ->
                %% Other error, we log
...
```

If concurrent execution of aux effects is implemented, the reverse is also true: the table could appear after that `ets:whereis/1` and before the insert/delete.

What do you think?

dumbbell: Ok, thank you for the feedback!

PR #296, opened 2024-09-11
Use `erlang:whereis/1` for checking if a store is running

the-mikedavis:
A store's process uses its `StoreId` as its registered name. This is a public interface of Ra so we can depend on it. Reading from key metrics counters is already very fast but switching to `whereis/1` eliminates basically all overhead of this function. When used heavily (for example in RabbitMQ while publishing and consuming rapidly) the CPU time spent on `is_store_running/1` disappears from the output of a perf recording and a flamegraph.

This is a follow-up to https://github.com/rabbitmq/khepri/pull/292 based on the post-merge discussion.

### DIFF
--- a/src/khepri_cluster.erl
+++ b/src/khepri_cluster.erl
@@ -1576,11 +1576,12 @@ get_store_ids() ->
 %% @doc Indicates if `StoreId' is running or not.
 
 is_store_running(StoreId) ->
-    ThisNode = node(),
-    RaServer = khepri_cluster:node_to_member(StoreId, ThisNode),
-    Timeout = khepri_app:get_default_timeout(),
-    KeyMetrics = ra:key_metrics(RaServer, Timeout),
-    Runs = maps:get(state, KeyMetrics) =/= noproc,
+    %% FIXME: Ra has no API to know if a Ra server is running or not. We could
+    %% use a public API such as `ra:key_metrics/2', but unfortunately, it is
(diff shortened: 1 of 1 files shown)


dumbbell (changes requested):
I still consider this is knowledge of an implementation detail of Ra.

I know that the `ra_server_id()` type is described as a registered process name in `src/ra.hrl`. However, it is explicitly documented as an internal thing:
* The Ra server ID concept is described in `docs/internals/INTERNALS.md`.
* When the `server_id()` type is defined and exported, it is also marked as internal:
    ```erlang
    %% export some internal types
    ...
    -type server_id() :: ra_server_id().
    ...
    ```

Therefore, I’m ok to accept the change, but I would like that a comment is added to clarify this. For instance:
```erlang
    %% FIXME: Ra has no API to know if a Ra server is running or not. We could
    %% use a public API such as `ra:key_metrics/2', but unfortunalety, it is
    %% not as efficient as querying the process directly. Therefore, we bypass
    %% Ra and rely on the fact that the server ID is internally a registered
    %% process name and resolve it to determine if it is running.
    Runs = erlang:whereis(StoreId) =/= undefined,
```

dumbbell (approved):
Thanks!

michaelklishin:
We don't have a `0.15.1` milestone (maybe we should) so I have set it to `0.16.0` because it's better than not having any details.

dumbbell:
Thank you, I forgot to set the milestone…

PR #297, opened 2024-09-23
Expire dedups on `tick`

the-mikedavis:
This change moves the work we do in `khepri_machine:drop_expired_dedups/2` out of `apply/3` and into the handler for Ra's periodic `tick` aux effect.

The commit message has more details but the TLDR is that the `maps:filter/2` within `drop_expired_dedups/2` can become expensive if many transactions are being handled before their dedups are received. Instead we can determine if any dedups are expired in the handler for the `tick` effect and expire any by committing a new `#expire_dedups{}` command.

Also included is a change to use the new `handle_aux/5` API introduced in Ra 2.10 but that change shouldn't have any effect on behavior.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -49,6 +49,8 @@
 %% <li>Changed the data structure for the reverse index used to track
 %% keep-while conditions to be a prefix tree (see {@link khepri_prefix_tree}).
 %% </li>
+%% <li>Moved the expiration of dedups to the `tick' aux effect (see {@link
+%% handle_aux/5}). This also introduces a new command `#drop_dedups{}'.</li>
 %% </ul>
 %% </td>
 %% </tr>
@@ -119,7 +121,8 @@
          get_dedups/1]).
 
 -ifdef(TEST).
--export([make_virgin_state/1,
+-export([do_process_sync_command/3,
+         make_virgin_state/1,
          convert_state/3,
          set_tree/2]).
 -endif.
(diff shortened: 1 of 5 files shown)


src/khepri_machine.erl
```diff
@@ -1252,16 +1255,30 @@ handle_aux(
...
-  AuxState, LogState, State) ->
+  AuxState, IntState) ->
+    State = ra_aux:machine_state(IntState),
     Tree = get_tree(State),
     ok = restore_projection(Projection, Tree, PathPattern),
-    {no_reply, AuxState, LogState};
-handle_aux(_RaState, _Type, _Command, AuxState, LogState, _MachineState) ->
-    {no_reply, AuxState, LogState}.
+    {no_reply, AuxState, IntState};
+handle_aux(_RaState, cast, tick, AuxState, IntState) ->
```
the-mikedavis, line 1267: @kjnilsson should this clause match on the leader state? Otherwise will we append the new command once for every member of the cluster?

src/khepri_machine.erl
```diff
@@ -1263,6 +1264,19 @@ handle_aux(
...
     {no_reply, AuxState, IntState};
+handle_aux(_RaState, cast, tick, AuxState, IntState) ->
+    State = ra_aux:machine_state(IntState),
+    Ts = erlang:system_time(millisecond),
+    Dedups = get_dedups(State),
+    CanExpire = fun(_CommandRef, {_Reply, Expiry}) -> Expiry =< Ts end,
+    Effects = case khepri_utils:maps_any(CanExpire, Dedups) of
+                  true ->
+                      ExpireDedups = #expire_dedups{},
+                      [{append, ExpireDedups}];
```
dumbbell, line 1275: What about collecting the lists of expired command refs during this scan and passing it in the `#expire_dedups{}` command? This way the apply only has to call `maps:remove/2`.

test/protect_against_dups_option.erl
```diff
@@ -127,62 +127,78 @@ dedup_and_dedup_ack_test() ->
...
+       {"Sleep and send the same transaction command",
+        ?_test(
+            begin
+                %% Sleep a little extra for the sake of slow CI runners.
+                %% During this sleep time the machine will receive Ra's
+                %% periodic `tick' aux effect which will trigger expiration of
+                %% dedups.
+                SleepTime = TickTimeout + erlang:floor(TickTimeout * 3 / 4),
+                timer:sleep(SleepTime),
+                %% The dedup should be expired so this command should
```
dumbbell, line 191: The comment looks to be truncated.

src/khepri_machine.erl
dumbbell: The state machine version was not bumped with this patch. An old machine that doesn’t know `#expire_dedups{}` will log an error and reply with a `?khepri_exception()`. Is that ok?

the-mikedavis: I thought it would be alright since I assumed followers drop the `mod_call` effect that logs an error but on second look followers execute those as well.

I have another patch coming that will need to bump the machine version anyways so I will group these changes together

the-mikedavis:
I think it should be easier to start with the other optimization (which also must bump the machine version) and this PR should depend on https://github.com/rabbitmq/ra/pull/474 so I'll mark this as a draft for now. We should be able to come back to it after the other PR is merged.

src/khepri_machine.erl
```diff
@@ -1266,16 +1271,48 @@ handle_aux(
...
-    {no_reply, AuxState, LogState}.
+    {no_reply, AuxState, IntState};
+handle_aux(leader, cast, tick, AuxState, IntState) ->
+    %% Expiring dedups in the tick handler is only available on versions 2
+    %% and greater. In versions 0 and 1, expiration of dedups is done in
+    %% `drop_expired_dedups/2'. This proved to be quite expensive when handling
+    %% a very large batch of transactions at once, so this expiration step was
+    %% moved to the `tick' handler in version 2.
+    case ra_aux:effective_machine_version(IntState) of
+        N when N >= 2 ->
```
dumbbell, line 1290: The variable is called `EffectiveMacVer` elsewhere. Could you please use the same name?

dumbbell (changes requested):

dumbbell (approved):

PR #298, opened 2024-09-24
Replace keep-while cond reverse index with a prefix tree

the-mikedavis:
When handling a deletion, `khepri_tree:eval_keep_while_conditions/2` folded over the reverse index of keep while conditions and evaluated any for which the removed path was a prefix. This meant folding over each non-leaf node in the tree. We can eliminate nodes very quickly though to limit the impact of this search by using a prefix tree instead.

This change switches out the `keep_while_conds_revidx` field in `#khepri_tree{}` to a new `khepri_prefix_tree` type when the Khepri cluster reaches the new effective machine version of 2.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -101,7 +101,6 @@
          get_tree/1,
          get_root/1,
          get_keep_while_conds/1,
-         get_keep_while_conds_revidx/1,
          get_triggers/1,
          get_emitted_triggers/1,
          get_projections/1,
@@ -156,7 +155,7 @@
 %% State machine's internal state record.
 -record(khepri_machine,
         {config = #config{} :: khepri_machine:machine_config(),
-         tree = #tree{} :: khepri_tree:tree(),
+         tree = khepri_tree:new() :: khepri_tree:tree(),
          triggers = #{} :: khepri_machine:triggers_map(),
          emitted_triggers = [] :: [khepri_machine:triggered()],
          projections = khepri_pattern_tree:empty() ::
@@ -2159,18 +2158,6 @@ get_root(State) ->
 get_keep_while_conds(State) ->
     #tree{keep_while_conds = KeepWhileConds} = get_tree(State),
     KeepWhileConds.
-
--spec get_keep_while_conds_revidx(State) -> KeepWhileCondsRevIdx when
-      State :: khepri_machine:state(),
-      KeepWhileCondsRevIdx :: khepri_tree:keep_while_conds_revidx().
-%% @doc Returns the `keep_while' conditions reverse index in the tree from the
-%% given state.
-%%
-%% @private
-
-get_keep_while_conds_revidx(State) ->
-    #tree{keep_while_conds_revidx = KeepWhileCondsRevIdx} = get_tree(State),
-    KeepWhileCondsRevIdx.
 
 -spec get_triggers(State) -> Triggers when
       State :: khepri_machine:state(),
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -2334,9 +2334,17 @@ make_virgin_state(Params) ->
 %%
 %% @private
 
-convert_state(State, MacVer, MacVer) ->
+convert_state(State, OldMacVer, NewMacVer) ->
+    lists:foldl(
+      fun(N, State1) ->
+              OldMacVer1 = N,
+              NewMacVer1 = erlang:min(N + 1, NewMacVer),
+              convert_state1(State1, OldMacVer1, NewMacVer1)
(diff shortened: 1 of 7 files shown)


the-mikedavis:
Dialyzer looks to get stuck in recursion on the Ubuntu + Erlang/OTP 27 runner, I can reproduce locally

dumbbell (changes requested):

src/khepri_machine.erl
```diff
@@ -166,10 +175,11 @@
          %% Added in machine version 1.
          dedups = #{} :: khepri_machine:dedups_map()}).
 
--opaque state_v1() :: #khepri_machine{}.
-%% State of this Ra state machine, version 1.
+-opaque state_v1_v2() :: #khepri_machine{}.
```
the-mikedavis, line 178: The dialyzer seems to be stuck if we write this as:

```erlang
-opaque state_v1() :: #khepri_machine{}.
-opaque state_v2() :: #khepri_machine{}.
-type state() :: state_v2() | state_v1() | khepri_machine_v0:state().
```

so as a compromise I've renamed the `state_v1` type into `state_v1_v2`

dumbbell: It would be more verbose, but I believe more correct if there was `tree_v0()` and `tree_v1()` that reference each the correct reverse index. For instance:

```erlang
% In khepri_tree.erl
-type tree_v0() :: #tree{keep_while_conds_revidx ::
                         khepri_tree:keep_while_conds_revidx_v0()}.
-type tree_v1() :: #tree{keep_while_conds_revidx ::
                         khepri_tree:keep_while_conds_revidx_v1()}.
-type tree() :: tree_v0() | tree_v1().
```

Then:

```erlang
% In khepri_machine.erl
-opaque state_v1() :: #khepri_machine{tree :: khepri_tree:tree_v0()}.
-opaque state_v2() :: #khepri_machine{tree :: khepri_tree:tree_v1()}.

-type state() :: state_v2() | state_v1() | khepri_machine_v0:state().
```

What do you think?

the-mikedavis: This is clearer to me :point_up: so I've made the change.

It turns out that the issue with the dialyzer seems to be having a union of opaque types, so if we switch `state_v1` and `state_v2` to be non-opaque, the dialyzer doesn't loop infinitely.

src/khepri_machine.erl
```diff
@@ -166,10 +175,13 @@
          %% Added in machine version 1.
          dedups = #{} :: khepri_machine:dedups_map()}).
 
--opaque state_v1() :: #khepri_machine{}.
+-type state_v1() :: #khepri_machine{tree :: khepri_tree:tree_v0()}.
 %% State of this Ra state machine, version 1.
 
--type state() :: state_v1() | khepri_machine_v0:state().
+-type state_v2() :: #khepri_machine{tree :: khepri_tree:tree_v1()}.
+%% State of this Ra state machine, version 2.
```
dumbbell, line 182: After putting more thoughts, I think we can use a single `state_v1()` opaque type. Both versions are covered by the underlying `tree_v0()` and `tree_v1()` types now.

src/khepri_machine_v0.erl
```diff
@@ -29,7 +29,7 @@
 
 -record(khepri_machine,
         {config = #config{} :: khepri_machine:machine_config(),
-         tree = #tree{} :: khepri_tree:tree(),
+         tree = khepri_tree:new() :: khepri_tree:tree(),
```
dumbbell, line 32: Here, the type should be `tree_v0()`.

src/khepri_tree.hrl
```diff
@@ -24,5 +24,4 @@
 
 -record(tree, {root = #node{} :: khepri_tree:tree_node(),
                keep_while_conds = #{} :: khepri_tree:keep_while_conds_map(),
-               keep_while_conds_revidx = #{} ::
-               khepri_tree:keep_while_conds_revidx()}).
+               keep_while_conds_revidx = #{}}).
```
dumbbell, line 27: Shouldn’t the type be left unmodified?

dumbbell: After testing to restore the type spec, Dialyzer complains because the type is opaque. I believe the complaint is valid. There is room for improvement in the type specs around `tree()` but it is out of scope for this pull request.

Meanwhile, can you use `-type` instead of `-opaque` for `khepri_tree:keep_while_conds_revidx()`?

dumbbell (commented):
This time, I tested my suggestions to make sure Dialyzer doesn’t loop indefinitely :)

dumbbell (approved):

PR #303, opened 2024-10-09
khepri_adv, khepri_tx_adv: Always return `khepri_adv:many_results()`

dumbbell:
## Why

Functions such as `khepri_adv:get()` and `khepri_adv:put()` extracted the single node's properties from the query/command return value and returned only that.

Therefore, the return value had a different form than other advanced functions' full node properties map.

In a future patch, we want to add the nodes deleted by the "expiration" of `keep_while` conditions. They can be deleted after a delete or a put. If we want to achieve that, we need to make all functions return the same thing.

## How

All functions that returned the single node props value are simplified to return the whole map, regardless of the number of tree nodes in the map.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -430,8 +430,7 @@
 %%
 %% `undefined' is returned if a tree node has no payload attached to it.
 
--type async_ret() :: khepri_adv:single_result() |
-                     khepri_adv:many_results() |
+-type async_ret() :: khepri_adv:many_results() |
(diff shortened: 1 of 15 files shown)


the-mikedavis (approved):
Just a minor typo I see in the second commit: "inser"=>"insert", otherwise I think this looks good :+1:

dumbbell:
> Just a minor typo I see in the second commit: "inser"=>"insert", otherwise I think this looks good 👍

Thank you! It is fixed.

PR #305, opened 2024-10-15
khepri_adv: Add `delete_reason` prop, accumulate keep-while expirations

the-mikedavis:
This change adds a `delete_reason` prop to the props map for nodes deleted with a function from the adv APIs `khepri_adv` and `khepri_adv_tx`. Tree nodes which are the target of deletion commands gain a `delete_reason` property set to `direct`. This change also includes any tree nodes which were deleted because of an expired keep-while condition in the node props map. These expired nodes have a `delete_reason` property set to `keep_while` instead. The `delete_reason` property lets callers filter on whether a change was a deletion and distinguish nodes deleted directly or indirectly.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -203,6 +203,22 @@
 -type child_list_length() :: non_neg_integer().
 %% Number of direct child nodes under a tree node.
 
+-type delete_reason() :: explicit | keep_while.
+%% The reason why a tree node was removed from the store.
+%%
+%% <ul>
+%% <li>`explicit' means that the tree node was removed from the store because a
+%% deletion command targeted that tree node or its ancestor. This reason is
+%% used when a tree node's path is provided to {@link khepri_adv:delete/3} for
+%% example.</li>
+%% <li>`keep_while': the tree node was removed because the keep-while condition
+%% set when the node was created became unsatisfied. Note that adding or
+%% updating a tree node can cause a keep-while condition to become unsatisfied,
+%% so a put operation may result in a tree node being deleted with this delete
+%% reason. See {@link khepri_condition:keep_while()} and {@link
+%% khepri:put_options()}.</li>
+%% </ul>
+
 -type node_props() ::
     #{data => khepri:data(),
       has_data => boolean(),
@@ -211,7 +227,8 @@
       payload_version => khepri:payload_version(),
       child_list_version => khepri:child_list_version(),
       child_list_length => khepri:child_list_length(),
-      child_names => [khepri_path:node_id()]}.
+      child_names => [khepri_path:node_id()],
+      delete_reason => khepri:delete_reason()}.
 %% Structure used to return properties, payload and child nodes for a specific
 %% tree node.
 %%
@@ -337,7 +354,8 @@
                                               child_names |
                                               payload |
                                               has_payload |
-                                              raw_payload],
+                                              raw_payload |
+                                              delete_reason],
                           include_root_props => boolean()}.
 %% Options used during tree traversal.
 %%
@@ -460,6 +478,7 @@
               payload_version/0,
               child_list_version/0,
               child_list_length/0,
+              delete_reason/0,
               node_props/0,
               trigger_id/0,
 
--- a/src/khepri_adv.erl
+++ b/src/khepri_adv.erl
@@ -332,11 +332,18 @@ put(StoreId, PathPattern, Data) ->
 %% in the path pattern is not met, an error is returned and the tree structure
 %% is not modified.
 %%
-%% The returned `{ok, NodeProps}' tuple contains a map with the properties and
-%% payload (if any) of the targeted tree node: the payload was the one before
-%% the update, other properties like the payload version correspond to the
-%% updated node. If the targeted tree node didn't exist, `NodeProps' will be
-%% an empty map.
+%% The returned `{ok, NodePropsMap}' tuple contains a map where keys correspond
+%% to the path to a node affected by the put operation. Each key points to a
+%% map containing the properties and prior payload (if any) of a tree node
+%% created, updated or deleted by the put operation. If the put results in the
+%% creation of a tree node this props map will be empty. If the put updates an
+%% existing tree node then the props map will contain the payload of the tree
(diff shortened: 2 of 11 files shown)


src/khepri.erl
```diff
@@ -203,6 +203,19 @@
 -type child_list_length() :: non_neg_integer().
 %% Number of direct child nodes under a tree node.
 
+-type delete_reason() :: direct | keep_while.
```
michaelklishin, line 206: I find `direct` to be a non-obvious reason. How about

``` erl
-type delete_reason() :: explicit | condition_lapsed.
```

I don't know if `condition_lapsed` is clearer than `keep_while` but `explicit` seems more obvious than `direct` to me.

Another option I have is

``` erl
-type delete_reason() :: user_action | condition_lapsed.
```

So, a key can be deleted as a result of a user action or a lapsed condition.

dumbbell: I like the `explicit` name.

`user_action` is unclear because it is a library, so there is no real users.

I prefer `keep_while` because it’s the name of the option to add the condition. Therefore the relation is explicit between the condition/option and the delete reason.

src/khepri_adv.erl
```diff
@@ -332,11 +332,17 @@ put(StoreId, PathPattern, Data) ->
...
-%% the update, other properties like the payload version correspond to the
-%% updated node. If the targeted tree node didn't exist, `NodeProps' will be
-%% an empty map.
+%% The returned `{ok, NodePropsMap}' tuple contains a map where keys correspond
+%% to the path to a node affected by the put operation. Each key points to a
+%% map containing the properties and prior payload (if any) of a tree node
+%% created, updated or deleted by the put operation. If the put results in the
+%% creation of a tree node this props map will be empty. If the put updates an
+%% existing tree node then the props map will contain the properties of the
+%% tree node before the update. The `NodePropsMap' map might also contain
```
dumbbell, line 341: This is not true for updates: the props map contains the new props, except for the payload.

src/khepri_adv.erl
```diff
@@ -452,11 +458,17 @@ put_many(StoreId, PathPattern, Data) ->
...
-%% then points to a map containing the properties and payload (if any) of the
-%% targeted tree node: the payload was the one before the update, other
-%% properties like the payload version correspond to the updated node.
+%% The returned `{ok, NodePropsMap}' tuple contains a map where keys correspond
+%% to the path to a node affected by the put operation. Each key points to a
+%% map containing the properties and prior payload (if any) of a tree node
+%% created, updated or deleted by the put operation. If the put results in the
+%% creation of a tree node this props map will be empty. If the put updates an
+%% existing tree node then the props map will contain the properties of the
+%% tree node before the update. The `NodePropsMap' map might also contain
```
dumbbell, line 467: Like above:
> This is not true for updates: the props map contains the new props, except for the payload.

src/khepri_adv.erl
```diff
@@ -925,18 +944,26 @@ delete_many(PathPattern, Options) when is_map(Options) ->
...
 %% {ok, #{[foo, bar] := #{data := value,
-%%                        payload_version := 1},
-%%        [baz, bar] := #{payload_version := 1}}} = khepri_adv:delete_many(
-%%                                                    StoreId, [foo, bar]).
+%%                        payload_version := 1,
+%%                        delete_reason := direct},
+%%        [baz, bar] := #{payload_version := 1,
+%%                        delete_reason := direct}}} =
+%% khepri_adv:delete_many(
+%%   StoreId, [?KHEPRI_WILDCARD_STAR, bar]).
```
dumbbell, line 966: This can be put on the same line.

src/khepri_tree.erl
```diff
@@ -528,20 +528,26 @@ find_matching_nodes_cb(_, {interrupted, _, _}, _, Acc, _) ->
...
     Node1 = remove_node_payload(Node),
     Node2 = remove_node_child_nodes(Node1),
-    NodeProps = gather_node_props(Node, TreeOptions),
-    {ok, Node2, Result#{Path => NodeProps}};
-delete_matching_nodes_cb(Path, #node{} = Node, TreeOptions, Result) ->
-    NodeProps = gather_node_props(Node, TreeOptions),
-    {ok, delete, Result#{Path => NodeProps}};
-delete_matching_nodes_cb(_, {interrupted, _, _}, _Options, Result) ->
+    NodeProps1 = gather_node_props(Node, TreeOptions),
+    NodeProps2 = NodeProps1#{delete_reason => DeleteReason},
```
dumbbell, line 542: This should honor the `props_to_return` tree option. The default list of props to return is defined by `?DEFAULT_PROPS_TO_RETURN` in `src/khepri_tree.hrl`.

src/khepri_tree.erl
```diff
@@ -528,20 +528,26 @@ find_matching_nodes_cb(_, {interrupted, _, _}, _, Acc, _) ->
...
+    {ok, Node2, Result#{Path => NodeProps2}};
+delete_matching_nodes_cb(
+  Path, #node{} = Node, TreeOptions, DeleteReason, Result) ->
+    NodeProps1 = gather_node_props(Node, TreeOptions),
+    NodeProps2 = NodeProps1#{delete_reason => DeleteReason},
```
dumbbell, line 547: Like above:
> This should honor the `props_to_return` tree option. The default list of props to return is defined by `?DEFAULT_PROPS_TO_RETURN` in `src/khepri_tree.hrl`.

src/khepri_tree.erl
dumbbell: I don’t know if this new property should be returned by default. I’m leaning towards no. What do you think?

dumbbell (changes requested):

src/khepri.erl
the-mikedavis: I went for

```erl
-type delete_reason() :: explicit | keep_while.
```

src/khepri_tree.erl
dumbbell: That’s a good point. Ok, let’s add it to the defaults.

dumbbell (approved):
Thank you!

PR #309, opened 2024-12-12
khepri_machine: Request `delete_reason` if effective machine version >= 2 only

dumbbell:
## Why

The `delete_reason` property to return was introduced with machine version 2. Older versions have two issues:
* They don't know this property
* They crash on unknown properties

## How

We request this new property only if the effective machine version is greater or equal to 2 to avoid the issues listed above.

Future unknown properties are ignored since commit 257cf328110a69618ad0cf35e3ea1c5bb2a40e4c merged in the previous pull request #308.

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -51,6 +51,9 @@
 %% </li>
 %% <li>Moved the expiration of dedups to the `tick' aux effect (see {@link
 %% handle_aux/5}). This also introduces a new command `#drop_dedups{}'.</li>
+%% <li>Added the `delete_reason' to the list of properties that can be
+%% returned. It is returned by default if the effective machine version is 2 or
+%% more.</li>
 %% </ul>
 %% </td>
 %% </tr>
@@ -98,8 +101,8 @@
 %% For internal use only.
 -export([clear_cache/1,
          ack_triggers_execution/2,
-         split_query_options/1,
-         split_command_options/1,
+         split_query_options/2,
+         split_command_options/2,
          split_put_options/1,
          insert_or_update_node/6,
          delete_matching_nodes/4,
@@ -118,7 +121,8 @@
          get_projections/1,
          has_projection/2,
          get_metrics/1,
-         get_dedups/1]).
+         get_dedups/1,
+         get_store_id/1]).
 
 -ifdef(TEST).
 -export([do_process_sync_command/3,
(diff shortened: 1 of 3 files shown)


dumbbell:
@the-mikedavis: As briefly stated in the commit message and a comment in the code, getting the effective machine version when handling transactions is a bit tricky. The code runs inside the Ra server process. It doesn’t have the store ID because we don’t keep it in the machine state, even though we get it in the init arguments…

For now, I just query the registered name of the process, but, yeah, not ideal… Ideally, I would modify the state to add the store ID. We already bump the machine version anyway. I was lazy for this first iteration :-)

So, I will prepare a new patch to modify that state, and I also need to document that the `delete_reason` was added as part of version 2.

Regardless, what do you think of the patch so far?

dumbbell:
I’m wrong: we have the store ID in the machine configuration!

src/khepri_machine.erl
```diff
@@ -51,6 +51,9 @@
 %% </li>
 %% <li>Moved the expiration of dedups to the `tick' aux effect (see {@link
 %% handle_aux/5}). This also introduces a new command `#drop_dedups{}'.</li>
+%% <li>Added the `delete_reason' to the list of properties that can be
+%% retured. It is returned by default if the effective machine version is 2 or
```
the-mikedavis, line 55:
```suggestion
%% returned. It is returned by default if the effective machine version is 2 or
```

src/khepri_machine.erl
```diff
@@ -1214,11 +1236,12 @@ can_skip_fence_preliminary_query(StoreId) ->
 
 -spec init(Params) -> State when
       Params :: machine_init_args(),
-      State :: state().
+      State :: khepri_machine_v0:state().
 %% @private
 
 init(Params) ->
-    %% Initialize the state.
+    %% Initialize the state. This function always return the oldest supported
```
the-mikedavis, line 1243:
```suggestion
    %% Initialize the state. This function always returns the oldest supported
```

the-mikedavis (commented):
I found a few typos but otherwise I think this looks great!

src/khepri_machine.erl
dumbbell: Fixed, thanks!

src/khepri_machine.erl
dumbbell: Fixed, thanks!

dumbbell:
@the-mikedavis: Did you check if mixed-version testing in RabbitMQ was fixed with this change?

the-mikedavis:
I didn't test this patch in the server but I was testing https://github.com/rabbitmq/khepri/commit/ba1bf1413f20d82a9ab44f8773c089bd2238ea85 before which should have roughly the same effect. We still need to handle the case of "paused" cluster members on the older version but that change did remove errors from `delete_reason` not being recognized on older members. I'll update https://github.com/rabbitmq/rabbitmq-server/pull/12753 to test this branch instead

the-mikedavis (approved):

dumbbell:
> I updated the draft PR in the server repo to use this change (and #312) and I don't see any error logs about delete_reason in the failed test logs 👍

Thank you! I couldn’t reproduce any errors related to this locally either.

PR #319, opened 2025-03-17
khepri_tx: Export the transaction API version

dumbbell:
## Why

The transaction code is compiled on one Erlang node with a specific version of Khepri. However, it is executed on all members of the Khepri cluster. Some Erlang nodes might use another version of Khepri, newer or older, and the transaction API may differ there.

For instance in Khepri 0.17.x, the return values of the `khepri_tx_adv` functions changed. The transaction code will have to handle voth versions of the API to work correctly. Thus it can use this function to adapt its behaviour.

## How

The first transaction version is set to 1 and is exported as a public API.

A transaction use can use it it determine how to use the `khepri_tx` and `khepri_tx_adv` modules and what to expect in terms of behaviour and return values.

### DIFF
--- a/src/khepri_tx.erl
+++ b/src/khepri_tx.erl
@@ -86,9 +86,12 @@
          clear_many_payloads/1, clear_many_payloads/2,
 
          abort/1,
-         is_transaction/0]).
+         is_transaction/0,
+         api_version/0]).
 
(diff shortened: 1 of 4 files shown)


dumbbell:
This change was reverted and superseded by #326.

PR #392, opened 2026-04-30
Revamp snapshot strategy

dumbbell:
The new snapshot strategy is based on the following two criterias:

1. Do not request snaspshots too often. By default, the last snasphot must at least 10 seconds old before the state machine considers requesting a new one. This time is configurable using the `snapshot_time_interval` state machine init argument. Note that a snapshot may not be requested after the configured interval: following criterias will be evaluated too. This step only ensures snapshots won’t be requested before the interval as elapsed.

2. Request a snapshot after unreleased command take more than 10 Mib of memory/disk space. The goal is to make sure a new snapshot will have a meaningful impact on the resources consumption. The threshold is configurable using the `unreleased_command_footprint_threshold` state machine init argument.

See the commits 68389a76d0d169cda5c634c102ab1835dffd8f4c and cdd07e733106bc15bfcc668e22f647b20bff80d3 in the branch. Other commits are refactors to prepare the ground and make eash commit easier to review.


dumbbell:
I'm not satisfied with the wrapping of commands inside `#with_annocations{}`. It's not future-proof enough. I'm going back to my initial idea of changing the command records to make them ~~extensible~~ easier to evolve.

src/khepri_machine.hrl
```diff
@@ -15,7 +15,7 @@
 %% The current/latest version of the state machine is defined in this macro
 %% instead of `khepri_machine:version/0' only. This way, it can be used in
 %% function specs too.
--define(LATEST_MACVER, 3).
+-define(LATEST_MACVER, 4).
 
 %% Map API behaviours with the state machine version they were introduced in.
 -define(API_BEHAV_MACVER_MAP,
```
mkuratczyk, line 21: I believe `command_annotations` is missing here

dumbbell: Good catch. This behaviour name is no longer correct and I should improve the spec of `api_behaviour()` to detect these errors.

dumbbell: Thank you, this is fixed.

PR #393, opened 2026-05-16
Allow overriding projection ETS option `protection`

ansd:
Some applications would like Khepri to own a projection ETS table, while still allowing other processes to insert and delete objects. Prior to this commit, this wasn't allowed since `protected` was hardcoded as default option.

### DIFF
--- a/src/khepri_projection.erl
+++ b/src/khepri_projection.erl
@@ -274,7 +274,8 @@ fun((Tids :: ets:tid() | #{atom() => ets:tid()},
                                      keypos => pos_integer(),
                                      read_concurrency => boolean(),
                                      write_concurrency => boolean() | auto,
-                                     compressed => boolean()}.
+                                     compressed => boolean(),
+                                     protection => ets:table_access()}.
 %% Overrideable ETS options.
 %%
 %% They can be used for single-table and multi-table projections. For the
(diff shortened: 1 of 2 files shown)


src/khepri_projection.erl
```diff
@@ -496,6 +498,9 @@ to_ets_options(compressed, true, Acc) ->
     [compressed | Acc];
 to_ets_options(compressed, false, Acc) ->
     Acc;
+to_ets_options(protection, P, Acc)
+  when P =:= public orelse P =:= protected orelse P =:= private ->
+    [P | [O || O <- Acc, O =/= public, O =/= protected, O =/= private]];
```
dumbbell: Could you please split that line into two statements: one that filters previous protections out, one that prepends a new one?

dumbbell (changes requested):

src/khepri_projection.erl
ansd: done

dumbbell (approved):

the-mikedavis (commented):
I'm sure there's a good use-case for this but a config option worries me since you can easily break the assumptions that Khepri has about whether records exist / don't exist, for simple projection functions. I will send a follow-up to add some warnings to the `options()` docs saying that you ought to know what you're doing

PR #398, opened 2026-05-30
khepri_tree: Don't crash in does_path_match when the matched node is absent

### LINKED ISSUE #405: does_path_match crash when a path-condition trigger is evaluated against a just-deleted node (0.17.2–0.18.0)
### What we ran into

We hit a state-machine crash in `khepri_tree:does_path_match/4` when a delete removes a node that a registered trigger's path pattern has to evaluate a *condition* against. `find_matching_nodes/3` returns `{error, {khepri, node_not_found, _}}` for the now-absent node, and the hard match

```erlang
{ok, #{CurrentPath := Node}} = find_matching_nodes(Tree, CurrentPath, TreeOptions)
```

badmatches inside `apply`.

The part that made it painful for us: the command is already committed to the Ra log, so it re-runs on every replay. `ra_server_sup` eventually reaches `reached_max_restart_intensity` and the store doesn't come back up on its own — and because the command replicates, all members are affected, not just one. We had to reset the Ra/Khepri state by hand to recover.

### Versions

We checked the released hex tarballs and the hard match is still present in **0.17.2 through 0.18.0**, so it looks like it affects all current releases.

### Roughly how to reproduce

1. Register a trigger whose path pattern includes a condition that has to inspect the node (e.g. `#if_data_matches{}`, `#if_has_payload{}`, `#if_child_list_length{}`).
2. Delete a node that trigger is evaluated against.
3. In the delete's aftermath the path-match queries the now-absent node → `node_not_found` → badmatch.

We came across it on an event-sourced store that uses Khepri triggers for change notifications. It seems specific to the trigger path — consumers that drive derived state via projections rather than triggers probably wouldn't reach it, so this may not affect the broker itself.

### Note on the fix

This is already covered by #398 (thanks @dumbbell for looking at it — it's been rebased onto `v0.17.x` as you suggested), with #400 as a related transaction error-propagation hardening. We're mostly opening this as a tracking issue so the impact is written down somewhere searchable.

No urgency implied, but since it can take a store down without a self-recovery path, it might be worth including #398 in a `v0.17.x` patch when convenient. Happy to add a regression test for the delete-triggers-condition case if that would help. Thanks for all the work on Khepri!

rgfaber:
## Summary

Defensive hardening of `khepri_tree:does_path_match/4` so it can never crash on
an absent node.

The path-condition clause queries the tree node at the current path to evaluate
the condition:

```erlang
{ok, #{CurrentPath := Node}} = find_matching_nodes(Tree, ..., TreeOptions),
```

If the matched path refers to a node that is **absent** from the tree,
`find_matching_nodes/3` returns `{error, {khepri, node_not_found, _}}` and the
`{ok, _} =` badmatch crashes the caller. This is especially severe when the
caller is the `khepri_machine` command apply loop: the badmatch aborts (and is
persisted with) the command, so it **replays and crashes the state machine on
every restart**, leaving the store unrecoverable.

## Relationship to 183a566

`183a566` ("khepri_machine: Use initial tree to list sprocs triggered by
deletions") already fixed the **known** way to reach this — deletion triggers
now evaluate against the pre-delete tree, where the node still exists.

This PR is complementary defense-in-depth: it makes `does_path_match/4` itself
total with respect to an absent node, so no current or future caller can
badmatch it into a persisted crash. A path condition that must query a node
which no longer exists cannot be met, so the path simply does not match (return
`false`). The post-lookup branch is extracted into `does_path_match_condition/7`
with no change to the matching logic.

Entirely reasonable to decline if you'd rather keep the assertion and rely on
callers always passing a tree in which the node exists.

## Test

`test/pattern_tree.erl` gains a direct `does_path_match/3` unit test that matches
a path absent from the tree using a path condition. Without this change it
reproduces the `node_not_found` badmatch; with it, `does_path_match/3` returns
`false`. `pattern_tree` 5/0, `triggers` 82/0 (incl. the 183a566 test), `xref`
clean.

## Background

Found in production: an event store built on Khepri/Ra scavenged (deleted)
streams that still had subscription triggers with path conditions registered,
hit this badmatch inside the state machine, and the persisted delete replayed
and bricked the store on every restart.

### DIFF
--- a/src/khepri_tree.erl
+++ b/src/khepri_tree.erl
@@ -899,22 +899,29 @@ does_path_match(
                                         payload_version,
                                         child_list_version,
                                         child_list_length]},
-    {ok, #{CurrentPath := Node}} = find_matching_nodes(
-                                     Tree,
-                                     lists:reverse([Component | ReversedPath]),
-                                     TreeOptions),
+    case find_matching_nodes(Tree, CurrentPath, TreeOptions) of
+        {ok, #{CurrentPath := Node}} ->
+            does_path_match_condition(
+              Component, Path, Condition, PathPattern, ReversedPath1, Tree,
+              Node);
+        {error, ?khepri_error(node_not_found, _)} ->
+            false
+    end.
+
+does_path_match_condition(
+  Component, Path, Condition, PathPattern, ReversedPath, Tree, Node) ->
     case khepri_condition:is_met(Condition, Component, Node) of
         true ->
             ConditionMatchesGrandchildren =
             case khepri_condition:applies_to_grandchildren(Condition) of
                 true ->
                     does_path_match(
-                      Path, [Condition | PathPattern], ReversedPath1, Tree);
+                      Path, [Condition | PathPattern], ReversedPath, Tree);
                 false ->
                     false
             end,
             ConditionMatchesGrandchildren orelse
-              does_path_match(Path, PathPattern, ReversedPath1, Tree);
+              does_path_match(Path, PathPattern, ReversedPath, Tree);
         {false, _} ->
             false
     end.
--- a/test/pattern_tree.erl
+++ b/test/pattern_tree.erl
@@ -128,6 +128,28 @@ fold_finds_all_patterns_matching_a_path_test() ->
       MatchingIndices([])),
     ok.
 
+does_path_match_tolerates_an_absent_node_test() ->
+    %% `does_path_match/3' must not crash when the matched path refers to a
+    %% node that is absent from the tree (e.g. matching the path of a node
+    %% that no longer exists). A path condition that has to query the
+    %% missing node cannot be met, so the path does not match.
+    %%
+    %% Previously the condition clause hard-matched
+    %% `{ok, _} = find_matching_nodes(...)'; for an absent node that lookup
+    %% returns `{error, {khepri, node_not_found, _}}', so the badmatch
+    %% crashed the caller. When the caller is the `khepri_machine' command
+    %% apply loop, the badmatch aborts (and is persisted in) the command,
+    %% so it replays on every restart.
+    TreePayload = #p_data{data = 1},
+    {ok, Tree, _, _} = khepri_tree:insert_or_update_node(
+                         khepri_tree:new(), [foo], TreePayload, #{}, #{}),
+    %% [foo, bar] is not in the tree; the condition at `bar''s position
+    %% forces a lookup of the absent node.
+    ?assertEqual(
+       false,
+       khepri_tree:does_path_match(
+         [foo, bar], [foo, #if_child_list_length{count = 0}], Tree)).
+
 %% Helper functions.
 
 -spec put_new(PatternTree, PathPattern, Payload) -> Ret when


dumbbell:
Thanks! I agree it makes sense that the `does_path_match/3` function returns a boolean if a path condition can’t be evaluated because the node is missing.

Could you please rebase your patch on the `v0.17.x` release branch, then change the target of this pull request?

rgfaber:
Done — rebased onto `v0.17.x` and retargeted the PR.

src/khepri_tree.erl
```diff
@@ -899,10 +899,25 @@ does_path_match(
...
-                                     lists:reverse([Component | ReversedPath]),
-                                     TreeOptions),
+    %% The node at `CurrentPath' may be absent — for example when matching
+    %% the path of a node that no longer exists in `Tree'. In that case
+    %% `find_matching_nodes/3' returns an error; a path condition that must
+    %% query the missing node cannot be met, so the path does not match.
+    %% (Returning `false' here rather than hard-matching `{ok, _} = ...'
+    %% avoids a badmatch: when this runs inside the `khepri_machine' command
+    %% apply loop, a badmatch aborts and is persisted with the command, so
+    %% it would replay and crash on every restart.)
```
dumbbell: Could you please remove this comment? It states the obvious and the historic part is in the commit message already where it belongs.

dumbbell (changes requested):

rgfaber:
Following up now that the rebase onto `v0.17.x` is in and the PR is retargeted.

In case it helps with prioritisation, some real-world impact: we hit this exact crash in production on an event-sourced store built on Khepri/Ra. A delete that fires a path-condition trigger evaluates `does_path_match` against the now-absent node; the hard match badmatches and crashes the state machine in `apply`. Because the offending command is already in the Ra log, it re-crashes on every replay, so `ra_server_sup` reaches `reached_max_restart_intensity` and the store stays down permanently — and since the command replicates, it takes down every replica, not just one. Recovery required wiping Ra state by hand.

The fix here (returning `false` when the node is absent) resolves it cleanly. Happy to add a regression test in whatever shape you'd prefer. Thanks for considering it!

src/khepri_tree.erl
```diff
@@ -899,10 +899,25 @@ does_path_match(
...
+    case find_matching_nodes(Tree, CurrentPath, TreeOptions) of
+        {ok, #{CurrentPath := Node}} ->
+            does_path_match_condition(
+              Condition, Component, Node, Path, PathPattern, ReversedPath1,
+              Tree);
+        {error, _} ->
```
dumbbell: In fact, could you please match specifically on the `node_not_found` error? Like:
```erlang
{error, ?khepri_error(node_not_found, _)} ->
```

dumbbell (changes requested):

dumbbell:
> In case it helps with prioritisation

I’m happy with the patch overall. But I shared a comment which is not addressed yet. I just shared another one that came to mind today.

rgfaber:
Apologies for the late turnaround — I was away on a short holiday.

Both comments addressed in the amended commit (force-pushed):

- Dropped the explanatory comment above the `case`; the rationale lives in the commit message.
- The error branch now matches specifically on `{error, ?khepri_error(node_not_found, _)}`, so any other error still fails loudly instead of being silently swallowed into a `false`.

The diff is now just the `case` plus the extracted `does_path_match_condition/7`.

`pattern_tree` 5/0, `triggers` 82/0, `xref` clean.

src/khepri_tree.erl
```diff
@@ -899,10 +899,17 @@ does_path_match(
...
+        {ok, #{CurrentPath := Node}} ->
+            does_path_match_condition(
+              Condition, Component, Node, Path, PathPattern, ReversedPath1,
+              Tree);
+        {error, ?khepri_error(node_not_found, _)} ->
+            false
+    end.
+
+does_path_match_condition(
+  Condition, Component, Node, Path, PathPattern, ReversedPath1, Tree) ->
```
dumbbell: I think this is my last nick pick :-)

Could you please:
* change the order of arguments to keep the one from the calling function, so `(Component, Path, Condition, PathPattern, ReversedPath, Tree)`?
* delete the `1` suffix to `ReversedPath1`?

Thank you!

dumbbell (changes requested):

src/khepri_tree.erl
rgfaber: Nitpicks very welcome, they make the patch better. Both done in the amended commit:

* Arguments now follow the calling function's order: `(Component, Path, Condition, PathPattern, ReversedPath, Tree)`. `Node` is the one extra the helper needs and your list didn't cover it, so I appended it last, which also matches `is_keep_while_condition_met_on_self/3` passing the node after the tree. Happy to move it if you'd rather have it elsewhere.
* The `1` suffix is gone from the helper's parameter. `ReversedPath1` stays in the caller only, where `ReversedPath` is already bound.

No logic change. `pattern_tree` 5/0, `triggers` 82/0, `xref` clean.

Thanks for the patient review!

dumbbell: Thank you! Good catch for the missing argument. I will let CI run, then I can merge your branch.

PR #399, opened 2026-06-10
khepri_projection: Add warning about overriding protection `ets` option

the-mikedavis:
This adds a small warning to the `overridable_ets_options()` type noting that inserting and deleting objects might break Khepri's assumptions about what's in a table.

References #393.


ansd (approved):
Thank you.

dumbbell (approved):

PR #401, opened 2026-06-23
Allow configuring `props_to_return` in sprocs

the-mikedavis:
With this change, you can configure `props_to_return` in the event filter options. This is mainly useful for giving the payload of the changed tree node to the registered stored procedure. For example:

    #{props_to_return => [payload]}

To introduce this change safely we need to bump the machine version to 4, and thread the machine version through the main insert_or_update_node and delete_matching_nodes helpers.

Closes #358.

### DIFF
--- a/src/khepri.erl
+++ b/src/khepri.erl
@@ -523,6 +523,8 @@
               favor_option/0,
               query_options/0,
               tree_options/0,
+              known_prop_to_return/0,
+              unknown_prop_to_return/0,
               put_options/0,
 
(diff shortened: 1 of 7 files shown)


the-mikedavis:
Looks like setup-beam must be broken within Windows runners at the moment

dumbbell:
I’m going to close this pull request because an alternative solution was committed with #425. We can discuss here the benefit of configuration that #425 doesn’t have.

PR #406, opened 2026-07-06
Optimize `keep_while` cascade deletes

mkuratczyk:
This PR optimizes a scenario where lots of nodes depend on a single node. Real-world examples include deleting a queue with lots of bindings (relatively rare scenario for RabbitMQ). For example:
```
make run-broker
wget https://github.com/rabbitmq/sample-configs/blob/main/topic-bindings/q0-with-100k-topic-bindings.json
rabbitmqctl import_definitions q0-with-100k-topic-bindings.json
rabbitmqctl delete_queue q0
```
The last command actually times out after 30 seconds without this PR on my machine, because Khepri needs ~150 seconds to process this deletion. With this PR, it completes in under 3 seconds.

For reference, flamegraphs captured at different stages of the deletion process:

khepri-2
khepri-1

### DIFF
--- a/src/khepri_tree.erl
+++ b/src/khepri_tree.erl
@@ -18,6 +18,7 @@
 -include("include/khepri.hrl").
 -include("src/khepri_error.hrl").
 -include("src/khepri_node.hrl").
+-include("src/khepri_tree.hrl").
 
 -export([new/0,
          get_root/1,
@@ -40,7 +41,8 @@
          convert_tree/3]).
 
 -ifdef(TEST).
--export([unopacify/1]).
+-export([unopacify/1,
+         siblings_chunks/2]).
 -endif.
 
 -record(tree, {root = #node{} :: khepri_tree:tree_node(),
@@ -1823,11 +1825,18 @@ paths_to_patterns([], PatternsToDelete) ->
                                   Path = ParentPath ++ [Component],
                                   [Path | Acc];
                               (ParentPath, Siblings, Acc) ->
-                                  %% Many siblings are deleted. We use a
-                                  %% single pattern that matches any siblings.
-                                  Pattern0 = #if_any{conditions = Siblings},
-                                  Pattern1 = ParentPath ++ [Pattern0],
-                                  [Pattern1 | Acc]
+                                  %% Many siblings are deleted. We use
+                                  %% patterns that match any siblings,
+                                  %% capped at `?MAX_SIBLINGS_PER_PATTERN'
+                                  Chunks = siblings_chunks(
+                                             Siblings,
+                                             ?MAX_SIBLINGS_PER_PATTERN),
+                                  lists:foldl(
+                                    fun(Chunk, Acc1) ->
+                                        Pattern0 = #if_any{conditions = Chunk},
+                                        Pattern1 = ParentPath ++ [Pattern0],
+                                        [Pattern1 | Acc1]
+                                    end, Acc, Chunks)
                           end, [], PatternsToDelete),
     PatternsToDelete1.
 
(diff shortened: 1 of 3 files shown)


michaelklishin (approved):
@mkuratczyk I have taken the liberty of fixing two minor issues as part of my review. The commit message explains them both.

src/khepri_tree.erl
```diff
@@ -1823,14 +1828,31 @@ paths_to_patterns([], PatternsToDelete) ->
...
+                                  %% Many siblings are deleted. We use
+                                  %% patterns that match any siblings,
+                                  %% capped at `?MAX_SIBLINGS_PER_PATTERN'
+                                  lists:foldl(
+                                    fun(Chunk, Acc1) ->
+                                        Pattern0 = #if_any{conditions = Chunk},
+                                        Pattern1 = ParentPath ++ [Pattern0],
+                                        [Pattern1 | Acc1]
+                                    end, Acc,
+                                    chunk(Siblings, ?MAX_SIBLINGS_PER_PATTERN))
```
dumbbell: Could you please split this into several statements? Like:
```erlang
Chunks = chunk(Siblings, ?MAX_SIBLINGS_PER_PATTERN),
lists:fold(
  fun(Chunk, Acc1) ->
          ...
  end, Acc, Chunks)
```

test/keep_while_conditions.erl
```diff
@@ -703,7 +707,7 @@ delete_node_with_many_dependents() ->
                      payload = khepri_payload:data(value)}],
     S0 = khepri_machine:init(?MACH_PARAMS(Commands)),
 
-    NumDependents = 1000,
+    NumDependents = 2500,
```
dumbbell: Could you please move the definition of `?MAX_SIBLINGS_PER_PATTERN` to `src/khepri_tree.hrl` and compute the value from it? Just in case we modify the constant in the future.

dumbbell (changes requested):

src/khepri_path.erl
```diff
@@ -605,7 +605,7 @@ component_targets_specific_siblings(#if_all{conditions = Conds}) ->
                   True ->
                       True;
                   {true, Siblings2} ->
-                      {true, Siblings1 ++ Siblings2};
+                      {true, Siblings2 ++ Siblings1};
```
dumbbell: This should be squashed into the appropriate commit.

src/khepri_path.erl
```diff
@@ -630,7 +630,7 @@ component_targets_specific_siblings(#if_any{conditions = Conds}) ->
...
```
dumbbell: This should be squashed into the appropriate commit.

src/khepri_tree.erl
```diff
@@ -1847,12 +1847,19 @@ paths_to_patterns([], PatternsToDelete) ->
...
+    {Chunk, Rest} = chunk_take(Count, List, []),
     [Chunk | chunk(Rest, Count)].
 
+chunk_take(0, Rest, Acc) ->
+    {lists:reverse(Acc), Rest};
+chunk_take(_Count, [], Acc) ->
+    {lists:reverse(Acc), []};
+chunk_take(Count, [Elem | Rest], Acc) ->
+    chunk_take(Count - 1, Rest, [Elem | Acc]).
+
```
dumbbell: If `lists:split/2` is to be somewhat copied to handle a list shorter than `Count`, let’s integrate this into the `chunk/2` function logic and make `chunk/2` tail-recursive.

The result should also be squashed into the appropriate commit.

dumbbell (changes requested):
Once all comments are addressed, the two resulting commits should be submitted as two separate pull requests.

dumbbell (approved):
The patch looks good, thank you!

I made a few cosmetic changes directly to the branch:
* Wrap a line crossing the 80-column boundary.
* Split nested calls and computations into several smaller expressions doing one thing, using intermediate variables to name each step.
* Change the copyright year in the new header to 2026 only.
* Delete the trailing empty line in the new header.

PR #407, opened 2026-07-08
khepri_path: Avoid quadratic sorting

mkuratczyk:
The `if_all` and `if_any` clauses re-sorted and deduplicated the accumulated sibling list on every fold step with
`lists:usort(Siblings1 ++ Siblings2)`. For a condition combining N sibling names (e.g. built by `khepri_tree:remove_expired_nodes/2` when a `keep_while` condition expiry cascades to many dependent nodes at once), this made the function O(N^2 log N) instead of O(N log N).

Defer the sort/dedup to a single pass after the fold instead of repeating it at every step.


dumbbell (approved):
Thank you!

I made a couple cosmetic changes, like reindenting the code to match the rest of the project.

PR #412, opened 2026-07-16
khepri_tree: Unwrap the `child_created` tag in `starting_node_in_rev_parent_tree/1`

mkuratczyk:
`#walk.reversed_parent_tree` holds `[#node{} | {#node{}, child_created}]` — the `child_created` tag marks an ancestor that was just created earlier in the same walk (see the `add_node_child`/`handle_branch` call sites that push it). `starting_node_in_rev_parent_tree/1` read the oldest entry off this list with plain `hd(lists:reverse(...))` without unwrapping that tag, so if the walk's outermost ancestor happened to be tagged, the function would return the `{#node{}, child_created}` tuple itself instead of the `#node{}` inside it. That value then flows into `Walk#walk{node = StartingNode}`, but `#walk.node`'s type is `#node{} | delete` — a tagged tuple there doesn't match any of the `walk_down_the_tree1/1` clauses that pattern-match on `#node{}` explicitly.

This mirrors the existing tag-unwrapping already done for the same field when popping a single parent for `?PARENT_KHEPRI_NODE` navigation (see the `case ParentNode0 of {PN, child_created} -> PN; _ -> ParentNode0 end` a few clauses above `walk_down_the_tree1/1`'s third clause).

### DIFF
--- a/src/khepri_tree.erl
+++ b/src/khepri_tree.erl
@@ -1278,15 +1278,18 @@ special_component_to_node_name(?THIS_KHEPRI_NODE, []) ->
 
 -spec starting_node_in_rev_parent_tree(ReversedParentTree) -> Node when
       Node :: tree_node(),
-      ReversedParentTree :: [Node].
+      ReversedParentTree :: [Node | {Node, child_created}].
 %% @private
 
 starting_node_in_rev_parent_tree(ReversedParentTree) ->
-    hd(lists:reverse(ReversedParentTree)).
+    case hd(lists:reverse(ReversedParentTree)) of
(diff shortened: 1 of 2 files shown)


dumbbell:
I prepared another testcase that triggers the bug from the Khepri API.

I would like to replace the unit test in `test/queries.erl` with this new one. Do you agree?

Also, because it’s a bug, the branch should be based on the v0.17.x release branch. Are you ok if I rebase it and update this pull request’s target?

mkuratczyk:
By all means, go ahead.

dumbbell (approved):
Thank you!

PR #413, opened 2026-07-17
khepri_condition: Wrap `false` in return value of is_valid/1

mkuratczyk:
`khepri_condition:is_valid/1` must always return `true` or `{fasle, Condition}`.

### DIFF
--- a/src/khepri_condition.erl
+++ b/src/khepri_condition.erl
@@ -721,18 +721,20 @@ compare_numerical_values(_, _)                 -> false.
 
 is_valid(Component) when ?IS_KHEPRI_PATH_COMPONENT(Component) ->
     true;
-is_valid(#if_node_exists{exists = Exists}) ->
-    is_boolean(Exists);
+is_valid(#if_node_exists{exists = Exists}) when is_boolean(Exists) ->
+    true;
 is_valid(#if_name_matches{}) ->
     true;
 is_valid(#if_path_matches{}) ->
     true;
-is_valid(#if_has_payload{has_payload = HasPayload}) ->
-    is_boolean(HasPayload);
-is_valid(#if_has_data{has_data = HasData}) ->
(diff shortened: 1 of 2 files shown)


src/khepri_condition.erl
```diff
@@ -721,18 +725,30 @@ compare_numerical_values(_, _)                 -> false.
 
 is_valid(Component) when ?IS_KHEPRI_PATH_COMPONENT(Component) ->
     true;
-is_valid(#if_node_exists{exists = Exists}) ->
-    is_boolean(Exists);
+is_valid(#if_node_exists{exists = Exists} = Cond) ->
+    case is_boolean(Exists) of
+        true  -> true;
+        false -> {false, Cond}
+    end;
```
dumbbell: This can be simplified as:
```erlang
is_valid(#if_node_exists{exists = Exists} = Cond)
  when is_boolean(Exists) ->
    true;
```

The last clause will catch the invalid case. Likewise for other checks.

src/khepri_condition.erl
```diff
@@ -35,6 +35,10 @@
 -include("include/khepri.hrl").
 -include("src/khepri_machine.hrl").
 
+%% is_valid/1's `false' branches guard against condition records whose
+%% boolean fields hold a non-boolean value at runtime
+-dialyzer({no_match, is_valid/1}).
```
dumbbell: Why do we need to disable this warning?

dumbbell (changes requested):

src/khepri_condition.erl
mkuratczyk: Seems like we don't after the suggested refactor.

src/khepri_condition.erl
mkuratczyk: I've pushed a refactored versio

PR #415, opened 2026-07-21
khepri_machine: reduce memory used when updating and restoring projections

mkuratczyk:
## Why

Refreshing a projection (`update_projection/4`) and restoring one after
recovery/snapshot (`restore_projection/3`) both fully materialized the
matching subtree into one or more `#{Path => NodeProps}` maps before
triggering the projection callback for each entry. On a large tree with
several registered projections, this showed up as a significant share of
the memory allocated while a node recovered its own state after a restart,
or while a follower caught up after a bulk write/delete on the leader.

## How

- `restore_projection/3` now folds directly over the tree and triggers the
  projection as each match is found, instead of collecting every match
  into a map first.
- `update_projection/4` now counts the matching nodes on both the old and
  new side (cheap), materializes whichever side is smaller into a map, and
  folds over the larger side, triggering the projection immediately for
  each match and removing it from the materialized map as it goes. This
  bounds peak memory to the smaller side regardless of which tree happens
  to be larger, instead of always fully materializing both sides (plus a
  diff of the two) as before.

Effectiveness tested using the attached tprof-based tests against `main` before rebasing on v0.17.x.

main:
```
restore_projection/3 (nodes=50000) allocated 9903121 words total (198.06 words/node, ceiling 190 words/node)
update_projection/4 (old=50000, new=500) allocated 12985157 words total (259.70 words/node of the larger side, ceiling 190 words/node)
update_projection/4 (old=50000, new=500) took 63635 us total (ceiling 175000 us)
update_projection/4 (old=500, new=50000) allocated 12985106 words total (259.70 words/node of the larger side, ceiling 190 words/node)
update_projection/4 (old=500, new=50000) took 71159 us total (ceiling 175000 us)
```

this PR:
```
reduce-projection-memory-usage:
restore_projection/3 (nodes=50000) allocated 6700668 words total (134.01 words/node, ceiling 190 words/node)
update_projection/4 (old=50000, new=500) allocated 8957554 words total (179.15 words/node of the larger side, ceiling 190 words/node)
update_projection/4 (old=50000, new=500) took 31300 us total (ceiling 175000 us)
update_projection/4 (old=500, new=50000) allocated 8957541 words total (179.15 words/node of the larger side, ceiling 190 words/node)
update_projection/4 (old=500, new=50000) took 35260 us total (ceiling 175000 us)
```

tprof-based tests used above:

[restore_projection_memory.erl.gz](https://github.com/user-attachments/files/30227150/restore_projection_memory.erl.gz)
[snapshot_projection_memory.erl.gz](https://github.com/user-attachments/files/30227151/snapshot_projection_memory.erl.gz)

### DIFF
--- a/src/khepri_machine.erl
+++ b/src/khepri_machine.erl
@@ -1531,12 +1531,13 @@ restore_projection(Projection, Tree, PathPattern) ->
     _ = khepri_projection:init(Projection),
     TreeOptions = #{props_to_return => ?PROJECTION_PROPS_TO_RETURN,
                     include_root_props => true},
-    case khepri_tree:find_matching_nodes(Tree, PathPattern, TreeOptions) of
-        {ok, MatchingNodes} ->
-            maps:foreach(fun(Path, Props) ->
-                                 khepri_projection:trigger(
-                                   Projection, Path, #{}, Props)
-                         end, MatchingNodes);
+    Fun = fun(Path, Props, Acc) ->
+                  khepri_projection:trigger(Projection, Path, #{}, Props),
+                  Acc
+          end,
+    case khepri_tree:fold(Tree, PathPattern, Fun, ok, TreeOptions) of
+        {ok, ok} ->
+            ok;
         Error ->
             ?LOG_DEBUG(
                "Failed to recover projection ~s due to an error: ~p",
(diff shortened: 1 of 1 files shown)


dumbbell (approved):
Thank you!

PR #436, opened 2026-09-17
khepri_machine: Make transaction functions "transparent" like stored procs

dumbbell:
## Why

Before this patch, the return value of `khepri:transaction()` was:
* `{ok, TxRet}` or
* an exception if the transaction function aborted or crashed

Any error, including with the communication with the Ra process (such as a `noproc` error or a timeout), was wrapped in `{ok, Error}` too.

A return value of `{ok, {error, Reason}}` is awkward at best. It also made its interpretation quite messy: is the error from Khepri or the transaction function itself?

This problem doesn't exist with an executed stored procedure because it is already transparent: it acts as if it was executed by the caller directly, like if Khepri was never involved.

Having the same behaviour with transaction fixes the interpretation of the return values and makes the global API more consistent.

## How

The return value of the transaction function is always wrapped in a `{txfun_ret, TxRet}` tuple internally. This helps distinguish it from other return values along the chain. Before returning anything to the caller, the transaction function return value `TxRet` is unwrapped.

Any errors coming from outside the transaction function are thrown as exception. For instance `error:{error, noproc}`.

`khepri:abort(Reason)` still raises a `throw:Reason` exception. That said, this API is deprecated because any Erlang exceptions will do the job just fine.

This is a breaking change in Khepri transaction API. Callers will have to be adapted because the return value is changed from `{ok, ActualRet}` to `ActualRet`.

The machine version is bumped to 5 and a new `transparent_tx_funs` behaviour is introduced to help with backward compatibility. Regardless of the effective machine version, the caller gets this new API.

References #400.


rgfaber:
Thanks, great rework and this is the API shape we were hoping for, indeed :)

I built `fa975f5` and ran the transaction suites (458 tests across simple/advanced tx, async_option, tx_sprocs, favor, dup-protection, db_info), all green. 

The three cases from my original PR are covered now: a stopped store raises 
`?khepri_error(tx_error, #{reason := {error, noproc}})`, a fun returning `{error, noproc}` comes back as-is, and `timeout => 0` raises. 

The result no longer depends on which machine version happens to be cached, which is exactly what our startup race needed.

On the reckon-db side the migration is small: we drop the `{ok, {error, E}} -> {error, E}` adaptations, match the value directly, and catch `tx_error` for the retry path. 
The structured `#{reason := ...}` gives us the pattern-matchable tri-state we asked for.

A few small observations, none blocking:

1. `fence/2` dropped the `{exception, ...}` clause from its `process_query` case, so a machine-side exception would now be returned as a raw `{exception, Class, Reason, Stack}` tuple instead of re-raised. The fence fun itself can't raise, so it's only reachable via a machine bug — but `fold/4` still re-raises. Is this Intentional asymmetry?

2. `khepri:async_ret()` doesn't cover the two new shapes `handle_async_ret/2` now produces: `{error, ?khepri_error(aborted_tx, ...)}` and `{error, {exception, _, _, _}}`. 
Maybe it's worth extending the spec so dialyzer consumers see them?

3. On a v4 machine, a tx fun legitimately returning `{error, noproc}` (or any whitelisted atom) gets raised as `tx_error`, an unavoidable ambiguity, but maybe worth a line in the release notes next to the breaking-change entry?

4. Is the whitelist (noproc | nodedown | shutdown | normal | timeout) the full set `ra:process_command/3` can return on the sync path? 
It's complete for `process_query` (noproc/timeout only), but I'm less sure about every ra error atom on the command path.

Happy to close #400 in favor of this once you're satisfied with it.

Thank you!

dumbbell:
> A few small observations, none blocking:
> 
> 1. `fence/2` dropped the `{exception, ...}` clause from its `process_query` case, so a machine-side exception would now be returned as a raw `{exception, Class, Reason, Stack}` tuple instead of re-raised. The fence fun itself can't raise, so it's only reachable via a machine bug — but `fold/4` still re-raises. Is this Intentional asymmetry?

The query function passed to `process_query/3` is responsible for catching and handling any exceptions. Most query functions do catch them and return the `{exception, Class, Reason, Stacktrace}` tuple. This is the case of the query function in `khepri_machine:fold()` (hidden in the patch), however this is not the case of the query function used in `khepri_machine:fence()`. Therefore it will never return the `exception` tuple. This was probably an copy-paste of the `fold()` code to create `fence()` and this was left.

> 2. `khepri:async_ret()` doesn't cover the two new shapes `handle_async_ret/2` now produces: `{error, ?khepri_error(aborted_tx, ...)}` and `{error, {exception, _, _, _}}`.
>        Maybe it's worth extending the spec so dialyzer consumers see them?

Good catch. In fact, this type is used in two incompatible ways: the return value of `handle_async_ret()` and the return value of asynchronous operations. The type is defined for the former case.

> 3. On a v4 machine, a tx fun legitimately returning `{error, noproc}` (or any whitelisted atom) gets raised as `tx_error`, an unavoidable ambiguity, but maybe worth a line in the release notes next to the breaking-change entry?

Good idea.

> 4. Is the whitelist (noproc | nodedown | shutdown | normal | timeout) the full set `ra:process_command/3` can return on the sync path?
>    It's complete for `process_query` (noproc/timeout only), but I'm less sure about every ra error atom on the command path.

I don’t know the full set of Ra errors. These are the errors I get from actual problems reaching a Ra server or from a terminating Ra server.

dumbbell:
> Good catch. In fact, this type is used in two incompatible ways: the return value of `handle_async_ret()` and the return value of asynchronous operations. The type is defined for the former case.

Scratch that. I got confused by `khepri_machine:async_ret()` vs. `khepri:async_ret()`.

